PR #3, opened 2023-10-27
Setup of Go Lambda Environment for AWS with SAM cli

### LINKED ISSUE #1: Initialize Go module and initial server to handle meetnearme backend
This issue is opened to build the basic Go lambda environment. The goal is to use AWS SAM Cli hosted inside a Docker container in order to isolate the Lambda environment and allow the synchronization with Docker Compose of other AWS services.

Brandon-G-Tripp:
# Description 

The PR establishes the baseline implementation of a HelloWorld AWS Lambda running in the Go runtime. It creates a simple APIProxyGateway that responds to requests at localhost:3001. The implementation serves to isolate the container run by AWS SAM cli within another docker container. The purpose of nesting the Docker containers is to integrate the Go Lambda API with docker compose. This feature will allow easier startup of future AWS backend services through the docker compose command. 

The api.Dockerfile was updated to include the Go runtime in the container for future testing. It also circumvents a volume mounting issue that arises with the combination of a build stage and docker compose volume of the PWD. The docker compose command will generate the appropriate Golang binary for the linux SAM container. 

Note: 

1. The local lambdas continue to run in sibling containers after tear down of main compose container. Run following command to kill all docker containers.

      `$ docker kill $(docker ps -q)`

2. A secondary pitfall to watch for, upon initial docker compose up, the binary will be generated in the root with the name `main`. If this executable is deleted by accident you will run into issues with it skipping the build. The solution is to create a blank `main` folder at root which docker will build against within the container. 

Fixes #1  

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

# Testing 

1. Pull changes into local branch
2. Make sure the Docker daemon is running (Docker Desktop)
3. run `$ docker compose build` from root directory
4. run `$ docker compose up` from root directory 
    - Screenshots below showing expected console output from successful run 
5. In another terminal window run `curl localhost:3001` - wait for response text from lambda 

### Successful build with `docker compose build`

Screen Shot 2023-10-27 at 11 47 55 AM

### Successful run with `docker compose up`

Screen Shot 2023-10-27 at 11 46 40 AM

### Successful endpoint hit with curl command

Screen Shot 2023-10-27 at 11 48 38 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 ## Getting Started
 
-1. `$ docker compose build`
-1. `$ docker compose up`
-1.
+1. `$ git clone git@github.com:meetnearme/api.git`
+2. `$ docker compose build`
+3. `$ docker compose up`
+4. `$ curl localhost:3001` to verify the server response
--- a/README.md
+++ b/README.md
@@ -1,1 +1,7 @@
-# api-go
+# Meet Near Me API
+
+## Getting Started
+
+1. `$ docker compose build`
+1. `$ docker compose up`
+1.
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -12,26 +12,12 @@ services:
     volumes:
       - $PWD:$PWD
       - /var/run/docker.sock:/var/run/docker.sock # this is for the nested container
-      # if we need to specificy mounting of .aws folder later, currently do not have
-      # - ~/.aws/:/root/.aws:ro
     working_dir: $PWD
     ports: 
       - "3001:3001"
     networks:
       - "backend"
     command: ["/bin/bash", "./sam_cli_entrypoint.sh"]
-    # links:
-    #   - "dynamodb-local"
-  # dynamodb-local:
-  #   command: "-jar DynamDBLocal.jar -sharedDB -dbPath ./data"
-  #   image: "amazon/dynamodb-local:latest"
-  #   container_name: dynamodb-local
-  #   ports: 
-  #     - "8000:8000"
-  #   volumes:
-  #     - "./docker/dynamodb:/home/dynamodblocal/data"
-  #   working_dir: /home/dynamodblocal
-
 
 networks: 
   backend:
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,27 @@
+
+version: '3.8'
+services:
+  sam: 
+    build: 
+      context: .
+      dockerfile: ./dockerfiles/api.Dockerfile
+    environment: 
+        - TZ= Etc/GMT
+        - SAM_CLI_TELEMETRY=false
+        - DOCKER_HOST=unix:///var/run/docker.sock
+    volumes:
+      - $PWD:$PWD
+      - /var/run/docker.sock:/var/run/docker.sock # this is for the nested container
+      - ~/.aws/:/root/.aws:ro
+    working_dir: $PWD
+    command: ["/bin/bash", "./sam_cli_entrypoint.sh"]
+    ports: 
+      - "3001:3001"
+    networks:
+      - "backend"
+
+
+networks: 
+  backend:
+    name: aws_backend
+    driver: bridge
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -12,13 +12,16 @@ services:
     volumes:
       - $PWD:$PWD
       - /var/run/docker.sock:/var/run/docker.sock # this is for the nested container
-      - ~/.aws/:/root/.aws:ro
+      # if we need to specificy mounting of .aws folder later, currently do not have
+      # - ~/.aws/:/root/.aws:ro
     working_dir: $PWD
-    command: ["/bin/bash", "./sam_cli_entrypoint.sh"]
     ports: 
       - "3001:3001"
     networks:
       - "backend"
+    command: ["/bin/bash", "./sam_cli_entrypoint.sh"]
+    # links:
+    #   - "dynamodb-local"
   # dynamodb-local:
   #   command: "-jar DynamDBLocal.jar -sharedDB -dbPath ./data"
   #   image: "amazon/dynamodb-local:latest"
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -19,15 +19,15 @@ services:
       - "3001:3001"
     networks:
       - "backend"
-  dynamodb-local:
-    command: "-jar DynamDBLocal.jar -sharedDB -dbPath ./data"
-    image: "amazon/dynamodb-local:latest"
-    container_name: dynamodb-local
-    ports: 
-      - "8000:8000"
-    volumes:
-      - "./docker/dynamodb:/home/dynamodblocal/data"
-    working_dir: /home/dynamodblocal
+  # dynamodb-local:
+  #   command: "-jar DynamDBLocal.jar -sharedDB -dbPath ./data"
+  #   image: "amazon/dynamodb-local:latest"
+  #   container_name: dynamodb-local
+  #   ports: 
+  #     - "8000:8000"
+  #   volumes:
+  #     - "./docker/dynamodb:/home/dynamodblocal/data"
+  #   working_dir: /home/dynamodblocal
 
 
 networks: 
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -19,6 +19,15 @@ services:
       - "3001:3001"
     networks:
       - "backend"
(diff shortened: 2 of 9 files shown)


dockerfiles/api.Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+# COPY go.mod go.mod
+# COPY go.sum .
+
+# RUN go mod download
+
+# COPY main.go .
+
+# RUN GOOS=linux go build -o main
+
```
brianfeister, line 14: @Brandon-G-Tripp Let's remove unused code and copy it to a feature branch somewhere rather than keep it in the `main` of the repo. Just ends up being "clutter" in many cases

docker-compose.yaml
```diff
@@ -0,0 +1,39 @@
...
+    #   - "dynamodb-local"
+  # dynamodb-local:
+  #   command: "-jar DynamDBLocal.jar -sharedDB -dbPath ./data"
+  #   image: "amazon/dynamodb-local:latest"
+  #   container_name: dynamodb-local
+  #   ports: 
+  #     - "8000:8000"
+  #   volumes:
+  #     - "./docker/dynamodb:/home/dynamodblocal/data"
+  #   working_dir: /home/dynamodblocal
```
brianfeister, line 33: Same here, delete and then just re-add in your feature branch for the dynamodb layer

README.md
```diff
@@ -1 +1,7 @@
-# api-go
\ No newline at end of file
+# Meet Near Me API
+
+## Getting Started
+
+1. `$ docker compose build`
+1. `$ docker compose up`
```
brianfeister, line 6: Add a validation step here along the lines of 

> Check that the server is running by curling it via`curl http://localhost:xyz "{ \"someJson\": \"hello\" }"`

dockerfiles/api.Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+ENV PATH /go/bin:$PATH
+
+RUN mkdir -p ${GOROOT}/src ${GOTPATH}/bin
+    
+RUN go version 
+
+# COPY --from=builder /build /app/build
+
+# Below is running command to hold open while docker compose works
+CMD [ "sh", "-c", "while sleep 3600; do:; done" ]
```
brianfeister, line 50: Interesting, did you say this is documented as a "known issue" somewhere? If so, let's add the URL for that issue here for future reference if this becomes invalid or irrelevant in the future

event.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key1": "value1",
+    "key2": "value2",
+    "key3": "value3"
+}
```
brianfeister, line 5: I think this should be deleted, right? Didn't end up being used

main.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import ( 
+	// "context"
```
brianfeister, line 4: delete if unused

template.yaml
```diff
@@ -0,0 +1,53 @@
...
+#   # https://github.com/awslabs/serverless-application-model/blob/master/docs/internals/generated_resources.rst#api
+#   HelloWorldApi:
+#     Description: "API Gateway endpoint URL for Prod stage for Hello World function"
+#     Value: !Sub "https://${ServerlessRestApi}.execute-api.${AWS::Region}.amazonaws.com/Prod/hello/"
+#   HelloWorldFunction:
+#     Description: "Hello World Lambda Function ARN"
+#     Value: !GetAtt HelloWorldFunction.Arn
+#   HelloWorldFunctionIamRole:
+#     Description: "Implicit IAM Role created for Hello World function"
+#     Value: !GetAtt HelloWorldFunctionRole.Arn
```
brianfeister, line 53: feel free to just store this somewhere in a feature branch if you'd like to keep it for reference, but let's not commit it to `main`

dockerfiles/api.Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+# install software commons 
+RUN sudo apt update \
+    && sudo apt install -y software-properties-common
+
+# isntall golang
+RUN sudo add-apt-repository ppa:longsleep/golang-backports -y \
+    && sudo apt update \ 
+    && sudo apt install -y golang-go
```
brianfeister, line 37: can we specify a version here? I feel like with `FROM ubuntu:focal` we have an implicit / non-declarative issue where we're not specifying go version explicitly but should be

dockerfiles/api.Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+ENV GOROOT /usr/lib/go
+ENV GOTPATH /go
+ENV PATH /go/bin:$PATH
+
+RUN mkdir -p ${GOROOT}/src ${GOTPATH}/bin
+    
+RUN go version 
+
+# COPY --from=builder /build /app/build
```
brianfeister, line 47: cleanup

sam_cli_entrypoint.sh
```diff
@@ -0,0 +1,2 @@
+GOOS=linux go build -o main
+sam local start-api -p 3001 --host=0.0.0.0 --container-host-interface=127.0.0.1 --container-host=host.docker.internal --docker-network aws_backend --warm-containers EAGER
```
brianfeister, line 2: 🏅 Nice, good to see you've got `--warm-containers` on by default. 

In the future we can consider an alternate entrypoint that intentionally simulates cold response, but it might not be relevant. I think I was reading that keeping 50 lambdas warm via AWS cron actions at a cost of around `$0.10 / mo`

go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/meetnearme/api
+
+go 1.20
+
+require github.com/aws/aws-lambda-go v1.41.0 // indirect
```
brianfeister, line 5: I was going to ask if this gets invoked on ever `docker compose up` but in testing I'm seeing it didn't get invoked after the first startup of the newly built container. 👍

docker-compose.yaml
```diff
@@ -0,0 +1,39 @@
...
+        - SAM_CLI_TELEMETRY=false
+        - DOCKER_HOST=unix:///var/run/docker.sock
+    volumes:
+      - $PWD:$PWD
+      - /var/run/docker.sock:/var/run/docker.sock # this is for the nested container
+      # if we need to specificy mounting of .aws folder later, currently do not have
+      # - ~/.aws/:/root/.aws:ro
+    working_dir: $PWD
+    ports: 
+      - "3001:3001"
```
brianfeister, line 19: When I start the server via `docker compose up` I see this

image

Any idea why it exposes both `http://172.18.0.2:3001` and `http://127.0.0.1:3001`? 

When I `curl http://172.18.0.2:3001` I get `curl: (56) Recv failure: Connection reset by peer`

But when I `curl http://127.0.0.1:3001` I get the expected `I am returning from the server`

Just wondering if we're mounting more processes than we realize through something implicit in the setup

brianfeister (commented):
Looks good, minor cleanup stuff and then I think we'll be good to go. 

Tested locally and it appears to work, this will be a huge win for collaborators using Windows and also people who just don't feel comfortable mucking with go binaries

Most importantly, you got it running at a functional level! Well done! 🙌  🚀

go.mod
Brandon-G-Tripp: Yes. SAM cli also does caching of the images that it builds internally so the Go code is compiled and started with SAM but sam will check for cached lambdas at runtime.

docker-compose.yaml
Brandon-G-Tripp: I will have to investigate more why it runs on the other couple of IP addresses also. I do know that it does not have to do with our specific setup with Docker and Docker Compose as the IPs are built when SAM cli invokes as a stand alone.

Brandon-G-Tripp:
Thank you for all the feedback @brianfeister. I believe everything is cleaned up and resolved. Go is pinned to 1.20 (the latest version). The dummy command was removed and tested and seems to be redundant.

docker-compose.yaml
brianfeister: Let's not worry about it for now. Not a blocker.

README.md
brianfeister: @Brandon-G-Tripp reminder: can you specify which port (since it exposes two, which is fine for now) users should `curl` and expect a response from?

brianfeister (approved):
Great work, disregard my recent comment that was on an older commit.

LGTM 🥇

PR #4, opened 2023-12-08
Local DyanmoDB Setup with Docker Compose 

Brandon-G-Tripp:
# Description 

The PR adds a Docker container to run the local implementation of DynamoDB. It aims to integrate a baseline dynamodb running locally and seeds the DB with basic event data. Also, it includes the API interface in Go to list all events in the Events table and to add new individual events. The changes mainly consist of moving from the mock user to using events as the data. 

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

# Testing 

1. Pull changes into local branch
2. Make sure the Docker daemon is running (Docker Desktop)
3. run `$ docker compose build` from root directory
4. run `$ docker compose up` from root directory 
    - Screenshots below showing expected console output from successful run 
5. In another terminal window run `curl localhost:3001/event` - wait for response text from lambda to list slice of events
6. Test Insert new event 

`curl -X POST -H 'Content-Type: application/json' -d '{"name": "Chess Tournament", "description": "Join the junior chess tournament to test your abilities", "datetime": "2024-03-13T15:07:00", "address": "15 Chess Street", "zip_code": "84322", "country": "USA"}' http://0.0.0.0:3001/event` 

7. Check insertion into list
`curl localhost:3001/event`


docker-compose.yaml
```diff
@@ -17,20 +16,47 @@ services:
     working_dir: $PWD
     ports: 
       - "3001:3001"
+    environment:
+      - AWS_ACCESS_KEY_ID=test
+      - AWS_SECRET_ACCESS_KEY=test
+      - AWS_SESSION_TOKEN=test
+      - REGION=us-east-1
```
brianfeister, line 23: do these need to pull from a `.env` file to be properly prepared for working in a deployed environment? As someone who deals with production applications with millions of users it gives me cringe to see these values as `test` even though my understanding is that this values only work on the local dockerized context. 

Ultimately I think we want the Docker image to pull the values from a file at the project root and that might be a `.env` file that we don't commit to version control for local and then we generate as a step in CI / CD by pulling from AWS Secrets Manager

README.md
```diff
@@ -5,3 +5,16 @@
 1. `$ docker compose build`
 1. `$ docker compose up`
 1.
+
```
brianfeister, line 8: Just pulled this branch and ran `docker compose build` and got this 

```
$ docker compose build
parsing /Users/bfeister/dev/meetnearme-api/docker-compose.yaml: yaml: unmarshal errors:
  line 19: mapping key "environment" already defined at line 7
```

bfeister: This line https://github.com/meetnearme/api/pull/4/files#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9L8-L11

and this line https://github.com/meetnearme/api/pull/4/files#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9R19-R23

Are duplicate keys, so the docker build fails... which is correct @Brandon-G-Tripp ?

Brandon-G-Tripp: Let me take a look at this. Somehow, the pull request lost my PR message. I had used the previous one as a template. I will check on all of these issues.

Brandon-G-Tripp: @bfeister those are actually not duplicate keys. One is for the Go api app and the other is for the temporary DB seeder service that is run in a separate container just to seed whatever we want from the seed files.

docker-compose.yaml
Brandon-G-Tripp: Yeah I agree that we should do that next step to make it ready for deployment. I think it will be easier now to go ahead and get the .env setup and include it in this PR. That way we already have a strategy for switching between dev and deploy and can have it ready as we build out our pipeline.

README.md
brianfeister: > @bfeister those are actually not duplicate keys. One is for the Go api app and the other is for the temporary DB seeder service that is run in a separate container just to seed whatever we want from the seed files.

I'm pretty sure they are duplicate keys. Have a look at this screenshot, both are immediate subkeys of the `sam` parent key (also my IDE confirms with a warning about dupe keys)

image

Once I merged them together, it was able to successfully build via `docker compose build` see below

image

main
brianfeister, line 1: @Brandon-G-Tripp is the binary meant to be committed (e.g. the file `main`) or just the golang src file `main.go`?

brianfeister (commented):
Good news is, once I made the change in this commit, I was able to get it up and running and hit the API successfully with the `POST` and the `GET` via `curl localhost:3001/event`

One thing aside from the bit about env vars I mentioned is REST verbs mean we should be naming this `/events` (plural) and not `/event`

template.yaml
```diff
@@ -12,42 +12,23 @@ Globals:
...
-      CodeUri: .
-      Timeout: 30
+      Timeout: 10
+      Handler: main 
       Events:
-        HelloWorld:
+        GetEvents:
+          Type: Api
+          Properties: 
+            Path: /event
```
brianfeister, line 24:
```suggestion
            Path: /events
```

template.yaml
```diff
@@ -12,42 +12,23 @@ Globals:
...
-#   HelloWorldApi:
-#     Description: "API Gateway endpoint URL for Prod stage for Hello World function"
-#     Value: !Sub "https://${ServerlessRestApi}.execute-api.${AWS::Region}.amazonaws.com/Prod/hello/"
-#   HelloWorldFunction:
-#     Description: "Hello World Lambda Function ARN"
-#     Value: !GetAtt HelloWorldFunction.Arn
-#   HelloWorldFunctionIamRole:
-#     Description: "Implicit IAM Role created for Hello World function"
-#     Value: !GetAtt HelloWorldFunctionRole.Arn
+            Path: /event
```
brianfeister, line 29:
```suggestion
            Path: /events
```

brianfeister (commented):
Here's a good guideline for REST verbs, they are typically plural, e.g.  `/events`

The api path for a single event tends to be an appended id, eg. `/events/:id`. 

So `GET /events/:id` gets the detail for just that event

and then `PATCH / PUT  /events/:id` would overwrite an existing event with new data

`POST /events` would be for adding a new event

image

4th answer down: https://stackoverflow.com/questions/6845772/should-i-use-singular-or-plural-name-convention-for-rest-resources

PR #5, opened 2024-02-02
Migration to sst htmx

brianfeister:
This PR converts the project to a full-stack HTMX + Go project (using `templ` for templating). 

It also includes @Brandon-G-Tripp 's work in setting up a dockerized SAM dynamodb environment from the previous (now refactored) PR here

Closes https://github.com/meetnearme/api/pull/4

image


.vscode/launch.json
Brandon-G-Tripp, line 1: @brianfeister Are these vscode configs supposed to be included in the checkin? I'm not sure from not having used vscode in a long time.

template.yaml
Brandon-G-Tripp, line 1: @brianfeister I am guessing at this point we should keep the template files for now even thought they are not in use because of the sst.config. We may need them for future.

Brandon-G-Tripp (approved):
Ok it looks good to me with the changes you finalized for SST. We may have some clean up but I feel we should leave the docker element for this merge for histories sake and then can factor it out later.

.vscode/launch.json
brianfeister: These are just SST default boilerplate. We could remove them, but seems like they can only help (or be deleted later if not)

template.yaml
brianfeister: Yeah I'll leave you to delete this one when you're ready. Let's try to remember. Also I'm fine with deleting now

brianfeister:
Since the Github Actions trigger is on `main` I need a merge to `main` to check if it works. If not, I'll follow with a fix PR

PR #8, opened 2024-02-14
Dynamodb prod wire up

brianfeister:
This PR does the following:

* Wire up prod + local dynamodb and make the code interoperable between the two
* Show a basic list of events on the home page, this event list is populated live from dynamoDb (total page response times feel like they're around ~500ms **including the DB read** which is 🔥🔥🔥)

## Video of it in action

Screen_Recording_2024-02-13_at_7 33 47 PM


brianfeister:
I think mainly it's just cleanup, there's a bunch of irrelevant code and
comments. I've been trying to get environment variables figured out but I
should do this in a follow up PR

brianfeister:
@Brandon-G-Tripp one thing that's new for me is the whole `myFunc( *var )` pass by reference rather than value thing. Can you keep me honest and check if I've missed any opportunities along those lines?

brianfeister:
@Brandon-G-Tripp I think this is ready to merge now. Let me know if anything in the Go code looks less than idiomatic

.github/workflows/deploy-prod.yml
```diff
@@ -1,29 +1,42 @@
...
+          ALLMYSECRETS: ${{ toJSON(secrets) }}
+          ALLMYVARS: ${{ toJSON(vars) }}
+        run: |
+          echo "MEETNEARME_TEST_SECRET=$(secrets.MEETNEARME_TEST_SECRET)" > .env
+        shell: bash
+      - name: 'print path'
+        run: |
+          echo "$(cat .env)"
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v2
```
Brandon-G-Tripp, line 35: Same with this aws-actions/ v4 was released a couple of weeks ago. I think we should bump it now before we get to far to see if we have any breaking changes.

functions/lambda/dynamodb.go
```diff
@@ -11,49 +12,66 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
 	"github.com/google/uuid"
+	"github.com/meetnearme/api/functions/lambda/shared"
 )
 
+func getDbTableName (tableName string) string {
+    var SST_Table_tableName_Events = os.Getenv("SST_Table_tableName_Events")
```
Brandon-G-Tripp, line 19: Just a question on this. Where do we load the .env?

Brandon-G-Tripp (changes requested):
Ok I do not seeing anything that is sticking out or does not seem idiomatic. I think it all looks good besides bumping those github actions before we merge.

.github/workflows/deploy-prod.yml
```diff
@@ -1,29 +1,42 @@
...
-              run: |
-                  npm i && npx sst deploy --stage prod
+  DeployProdSST:
+    runs-on: ubuntu-latest
+    env:
+      #Define your envs needed for static generation:
+      MEETNEARME_TEST_SECRET: ${{ secrets.MEETNEARME_TEST_SECRET }}
+    steps:
+      - name: Git clone the repository
+        uses: actions/checkout@v3
```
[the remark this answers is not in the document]
brianfeister: Good catch, fixing now

.github/workflows/deploy-prod.yml
brianfeister: Good call, pushing a fix to test if anything breaks

functions/lambda/dynamodb.go
brianfeister: The `.env` is a mistake and I removed it from this PR. The call to it will appear in my next PR that introduces env vars (very important!) 

I don't want this PR to add it now, as it would create confusion when looking at the upcoming PR, there is no clear addition of `.env` to `.gitignore` which would be weird

brianfeister:
@Brandon-G-Tripp resolved all of your comments and the deployment run passes with the updated versions of the github actions modules

functions/lambda/dynamodb.go
Brandon-G-Tripp: Perfect makes sense. Just wanted to clarify because the Getenv is a bit tricky. I have been using godotenv package for other projects and seems to be the main package in use in the go ecosystem.

Brandon-G-Tripp (approved):
It looks good to me to move onto the env vars and next features.

PR #9, opened 2024-02-22
Feature/env vars

brianfeister:
This PR adds support for environment variables throughout the Go app. Note there are multiple places where changes are required. I spent significant time trying to reduce this to one place, but in the end it just wasn't possible.

On the up side, the reusable env vars github action `.github/actions/set_aws_creds_env_vars/action.yml` adds some assurance via it's declared dependencies that any workkflow that uses it as a step (we will have `prod` and `feature/*` deployments at minimum) will fail if not explicitly passing the full allow list of env vars. These are [declared here]( https://github.com/meetnearme/api/compare/main...feature/env-vars#diff-a6c9404904fd30e5a1d76c22785a0e7c4a99b2db492105f29241b9fa44ec8a19R3-R8) and [passed here](https://github.com/meetnearme/api/compare/main...feature/env-vars#diff-4ce342fb97e7d7feb8a578c91ba89fe4c27ddf16b4306909fe6d90220141913dR24-R25).

The other required place to explicitly add them is [here in `stacks/ApiStack.ts`](https://github.com/meetnearme/api/compare/main...feature/env-vars#diff-324f081e8d04fa75b7673e77305af862edd739283d27077a060def0f55c755e5R12-R15)

image


.github/workflows/deploy-prod.yml
```diff
@@ -15,28 +15,20 @@ permissions:
...
-      - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@v4
-        with:
-          role-to-assume: arn:aws:iam::626762246796:role/GitHub
-          role-duration-seconds: 1200 #adjust as needed for your build time
-          aws-region: us-east-1
+          echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
+        id: extract_branch
       - name: Deploy app
+        # npm i && npx sst deploy --stage $stage-${{ steps.extract_branch.outputs.branch }}
```
brianfeister, line 32: will fix in follow up PR that will add support for deploying `feature/*` branches that don't overwrite prod SST stage

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 .sst/*
+.env
```
brianfeister, line 3: need to add `.env.example` in repo so folks know they need an untracked `.env`

functions/lambda/main.go
```diff
@@ -49,7 +50,10 @@ func Router(ctx context.Context, req events.APIGatewayV2HTTPRequest) (events.API
         if eventsErr != nil {
             return serverError(eventsErr)
         }
-        component := views.Home(Pages, eventList)
+
+        meetNearMeTestSecret := os.Getenv("MEETNEARME_TEST_SECRET")
+        component := views.Home(Pages, eventList, meetNearMeTestSecret)
+
```
brianfeister, line 56: want to keep some example of usage, will delete when we integrate for real

sst.config.ts
```diff
@@ -12,9 +12,13 @@ export default {
...
   },
   stacks(app) {
+    // oddly, order matters here, don't switch the order
     app.setDefaultFunctionProps({
       runtime: 'go',
     });
     app.stack(StorageStack).stack(ApiStack);
+    app.addDefaultFunctionPermissions({
+      permissions: ['dynamodb:*'],
+    });
```
brianfeister, line 22: should make dynamo permitted from all functions, including those outside the API gateway

.github/actions/set_aws_creds_env_vars/action.yml
```diff
@@ -0,0 +1,25 @@
...
+  MEETNEARME_TEST_SECRET:
+    description: 'The secret for the MeetNearMe test environment'
+    required: true
+  ZENROWS_API_KEY:
+    description: 'The API key for the ZenRows environment'
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Make envfile
```
j4inam, line 12: We can directly run shell commands and access Github Actions variables in the workflow files.
While nothing wrong in using a third party action, we could simply do:

```
name: Create .env file
        run: |-
          touch .env
          echo MEET_NEARME_TEST_SECRET=${{ secrets. MEETNEARME_TEST_SECRET }} >> .env
          echo ZENROWS_API_KEY =${{ secrets. ZENROWS_API_KEY }} >> .env
          cat .env
```

Just one less dependency to worry about. :p

j4inam (approved):

Brandon-G-Tripp (commented):
Looks good to me. Can we add the three locations that the .env secrets are needed to the README for documentation @brianfeister ?

PR #13, opened 2024-02-26
Feature: Deploy feature branch previews

brianfeister:
This PR adds `.github/workflows/deploy-feature.yml` which enables deployment via github actions to an AWS namespace `feature-*` for any branch that begins with `feature/*`. This will accelerate cross-workstream validation as each PR should get it's own AWS deployment.

Right now, this means that a new DynamoDB instance is stood up for each `feature/*` branch. This is not optimal, as we _most likely_ want a shared pre-production database with all of our test data. Another option would be to seed DynamoDB in a creation step so that each `feature-*` aws cloudformation deployment comes with the needed array of test data for reasonable manual validation. ([Discussion on pre-production data strategy here](https://github.com/meetnearme/api/issues/14))

But pre-production database decisions are for another day. For now, we will allow the creation of an empty database, and say that the empty state of the events list here is correct:

image

The AWS / SST stack deployment specific to this branch can be found here:
https://xsc64a8m34.execute-api.us-east-1.amazonaws.com/

And data could be pushed to this DB instance via a POST CURL to https://xsc64a8m34.execute-api.us-east-1.amazonaws.com/ if desired

The Github Actions CI run that deployed this is here:
https://github.com/meetnearme/api/actions/runs/8049963412/job/21984511133

**Follow-up / TODO:**
- [ ] https://github.com/meetnearme/api/issues/15
- [ ] https://github.com/meetnearme/api/issues/14


.github/actions/set_aws_creds_env_vars/action.yml
```diff
@@ -6,15 +6,19 @@ inputs:
   ZENROWS_API_KEY:
     description: 'The API key for the ZenRows environment'
     required: true
+  GIT_BRANCH_NAME:
+    description: 'The name of the branch being deployed'
```
brianfeister, line 10: we want this all the way through actual deployment as we will use this in `sst.config.ts` and it needs to be present at deploy time

.github/workflows/deploy-feature.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  DeployFeatureSST:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git clone the repository
+        uses: actions/checkout@v4
+      - name: Extract branch name
+        shell: bash
+        # sed 's|/|-|g' replaces all slashes with hyphens, aws stack names can't have slashes
```
brianfeister, line 23: does what it says, `sed` to replace `/` characters was necessary for this to work (otherwise it was throwing weird / misleading errors that weren't the real problem... real problem is AWS doesn't allow `/` in CloudFormation stack names)

.github/actions/set_aws_creds_env_vars/action.yml
```diff
@@ -6,15 +6,19 @@ inputs:
...
+    required: false
```
brianfeister, line 11: not `required` because `.github/workflows/deploy-prod.yml` expects this not to be set

.github/workflows/deploy-feature.yml
```diff
@@ -0,0 +1,35 @@
+name: Deploy Feature Branch (SST)
+on:
+  push:
+    branches:
+      - feature/*
```
j4inam, line 5: Smart move. I was about to ask how do you plan to enforce branch-naming. This will work :D

j4inam (approved):

j4inam:
LGTM 🚀

brianfeister:
Here we go, `feature/*` preview deployments incoming

PR #16, opened 2024-02-27
Update SST setup instructions in the readme

j4inam:
For devs setting up this repo for the first time, the instructions to link SST using AWS is missing important steps.
This PR adds instructions directly to the README and updates .env.example to be in-line with the instructions

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,4 @@
 MEETNEARME_TEST_SECRET=anything
 ZENROWS_API_KEY=ask_for_key
+aws_access_key_id="<YOUR_AWS_IAM_ACCESS_KEY_ID>"
+aws_secret_access_key="<YOUR_AWS_IAM_ACCESS_KEY_SECRET>"
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,2 @@
 MEETNEARME_TEST_SECRET=anything
 ZENROWS_API_KEY=ask_for_key
-aws_access_key_id="<YOUR_AWS_IAM_ACCESS_KEY_ID>"
-aws_secret_access_key="<YOUR_AWS_IAM_ACCESS_KEY_SECRET>"
--- a/README.md
+++ b/README.md
@@ -11,18 +11,8 @@
 
 1. `npm i`
 1. Create an AWS account if you don't have one.
-1. Create an IAM Role: `meetnearme-bot`.
-1. [Optional] Assign IAM User to User group: `meetnearme`
-1. Under the IAM Role > Select the **Permissions** Tab
-1. Select _Attach existing policies directly_.
-1. Search for **AdministratorAccess** and select the policy by checking the checkbox, then select **Next**.
-1. Click **Add Permissions**
-1. Navigate back to the IAM Role
-1. Go to **Security Credentials** Tab
-1. Select **Create access key**
-1. Select **Other** and select **Next**
-1. Optionally add a tag and select **Create Access Key**
-1. Create an `.env` file using `.env.example` to add needed keys.
+1. [Create an IAM User](https://sst.dev/chapters/create-an-iam-user.html)
+1. Export `aws_access_key_id` and `aws_secret_access_key` env variables.
 1. Run `brew install awscli` in the terminal to install AWS CLI
 1. Run `aws configure` to [Authorize SST via AWS CLI](https://sst.dev/chapters/configure-the-aws-cli.html)
 17. Run `npm run dev` to run the Go Lambda Gateway V2 server locally, proxied through
--- a/README.md
+++ b/README.md
@@ -5,26 +5,26 @@
 ### Running the local SAM dynamodb docker container
 
 1. `$ docker compose build`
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -26,8 +37,8 @@ export GOPATH=$HOME/go
 export PATH=$PATH:$GOROOT/bin:$GOPATH/bin
 ```
 
-1. `go install github.com/a-h/templ`
-1. Run `templ generate`
+2. `go install github.com/a-h/templ`
+3. Run `templ generate`
```
brianfeister, line 41: `1` and then `2` here right?

README.md
```diff
@@ -5,16 +5,27 @@
...
+8. Click **Add Permissions**
+9. Navigate back to the IAM Role
+10. Go to **Security Credentials** Tab
+11. Select **Create access key**
+12. Select **Other** and select **Next**
+13. Optionally add a tag and select **Create Access Key**
+14. Create an `.env` file using `.env.example` to add needed keys.
+15. Run `brew install awscli` in the terminal to install AWS CLI
+16. Run `aws configure` to [Authorize SST via AWS CLI](https://sst.dev/chapters/configure-the-aws-cli.html)
+17. Run `npm run dev` to run the Go Lambda Gateway V2 server locally, proxied through
```
brianfeister, line 28: The reason I use `1. <step description>` is because it's a markdown hack to save you from worrying about re-numbering everything when adding steps. Notice that my current `1. ` repeated, renders like this on github (here for example https://github.com/meetnearme/api?tab=readme-ov-file#running-the-lambda-project):

image

.env.example
```diff
@@ -1,2 +1,4 @@
 MEETNEARME_TEST_SECRET=anything
 ZENROWS_API_KEY=ask_for_key
+aws_access_key_id="<YOUR_AWS_IAM_ACCESS_KEY_ID>"
+aws_secret_access_key="<YOUR_AWS_IAM_ACCESS_KEY_SECRET>"
```
brianfeister, line 4: let's discuss this one, I ran `awscli` in a way where these keys are stored in my `~/` directory and not in `.env`. Not sure which is correct

README.md
j4inam: Ohhh TIL
I did find it weird as to why was everything numbered 1 lol. I’ll revert that back

.env.example
brianfeister: Also, I want `.env` to be stuff that lines up with what we ship to the lambda in it's `.env` file, we wouldn't ship these as part of `.env` in deployment

README.md
j4inam: 1 is above this. I reverted back to auto-numbering.

README.md
j4inam: Reverted back to auto-numbering :p

.env.example
j4inam: Yeah kinda makes sense not having this in .env since this is more of a local-dev thing. 
We don't actually need these for the app to run I guess

README.md
```diff
@@ -10,11 +10,22 @@
...
+1. Click **Add Permissions**
+1. Navigate back to the IAM Role
+1. Go to **Security Credentials** Tab
+1. Select **Create access key**
+1. Select **Other** and select **Next**
+1. Optionally add a tag and select **Create Access Key**
+1. Create an `.env` file using `.env.example` to add needed keys.
+1. Run `brew install awscli` in the terminal to install AWS CLI
+1. Run `aws configure` to [Authorize SST via AWS CLI](https://sst.dev/chapters/configure-the-aws-cli.html)
+17. Run `npm run dev` to run the Go Lambda Gateway V2 server locally, proxied through
```
brianfeister, line 28: Higher level question: are these steps copied from the SST docs? Do we feel there's a risk that SST might update their docs and then we would have outdated steps without realizing that problem?

I was imagining just linking to the various SST / AWS docs that are linked out from that initial https://sst.dev/chapters/configure-the-aws-cli.html but I'm open to being wrong about this

j4inam: > 1. Under the IAM Role > Select the **Permissions** Tab
> 1. Select _Attach existing policies directly_.
> 1. Search for **AdministratorAccess** and select the policy by checking the checkbox, then select **Next**.
> 1. Click **Add Permissions**
> 1. Navigate back to the IAM Role
> 1. Go to **Security Credentials** Tab
> 1. Select **Create access key**
> 1. Select **Other** and select **Next**
> 1. Optionally add a tag and select **Create Access Key**

This block is generic instructions to create an IAM user in AWS for any use-case. It's not SST specific.

I don't mind switching to SST doc links though.

brianfeister: I think the SST docs link out to the AWS instructions on this, I believe this might the link we want:

https://docs.sst.dev/advanced/iam-credentials

brianfeister: Also this one:
https://sst.dev/chapters/create-an-iam-user.html

brianfeister (approved):
Thanks for adding these, my mistake leaving them out. LGTM 🚀

PR #17, opened 2024-02-27
Feature: Enable auto-deletion of SST / AWS CloudFormation stacks when the feature branch merges

brianfeister:
Add `.github/workflows/remove-feature.yml` to enable auto-deletion of SST / AWS CloudFormation stacks that begin with `feature-*` when a branch of that name is deleted (a side effect of merge)

Note this is one of those PRs that has to merge and be in `main` in order for the `on: delete: branches: 'feature/**' to work. It could get messy if I don't know what I'm doing, but I think it might be lined up to work properly with what I have here already

### DIFF
--- a/.github/workflows/remove-feature.yml
+++ b/.github/workflows/remove-feature.yml
@@ -1,0 +1,30 @@
+name: Remove Feature Branch (SST)
+on:
+  delete:
+    branches:
+      - 'feature/**'
+jobs:
+  DeleteFeatureSST:
+    if:
+      github.event.ref_type == 'branch' && startsWith(github.ref,
+      'refs/heads/feature/')
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/remove-feature.yml
```diff
@@ -0,0 +1,30 @@
+name: Remove Feature Branch (SST)
+on:
+  delete:
```
j4inam, line 3: Do we want to trigger this on branch merge ?
If I merge and don't hit delete branch the stack will stick around.

j4inam (approved):

.github/workflows/remove-feature.yml
brianfeister: In the case of this repo, I've enabled "auto-delete branches on merge" at the github settings level so we don't forget that step

PR #24, opened 2024-02-28
add missing permissions for AWS OIDC to SST remove feature workflow

brianfeister:
self-merging, iteration is the only way in the absence of github actions debugging methods


brianfeister:
This PR finally succeeded in running `sst remove <feature/*>` and the successful deletion job can be found here:

https://github.com/meetnearme/api/actions/runs/8076183187/job/22064242348

PR #25, opened 2024-02-29
Integrate gin, htmx and tailwind with API Gateway

j4inam:


brianfeister:
Closing in favor of the hand-rolled / non-gin framework you proposed in #31 @j4inam

PR #26, opened 2024-03-04
Feature: Tailwind w static assets, dev tooling

brianfeister:
This PR brings together a bunch of build processes that we need in order to run the whole project end-to-end locally (and in some cases, missing elements we need for production like a CloudFront + S3 static asset strategy)

- [x] Add CloudFront + S3 deployments of `static` directory in `stacks/StaticSiteStack.ts`, wire this up to frontend (this is broken in deployments because Amazon needs to "verify my account" but the deployment works from local dev)
- [x] Add daisy UI + tailwind + tailwind watcher (also, to prevent "I forgot to build the CSS, whoops!" errors, the `static/assets/styles.css` file generated by tailwind is ignored in `gitignore` and every deployment re-runs the `tailwind:prod` command to generate minified CSS from source
- [x] Adds a new series of commands (and scripts where they're needed) to unify (and prevent "gotchas" with the developer tooling) and concurrently run all of these processes: 
    1. tailwind generation
    2. templ generation
    3. dynambodb local via SAM
    4. sst dev (for live lambda development)
    5. serve-static (for locally serving tailwind files without the need to have a manual restart of `sst dev` in order to push the latest CSS to S3)
    These start the processes concurrently, with two operating modes: 1) `npm run dev` - this starts all of the processes in a single terminal session, writing to stdout from each of them in a way that interleaves, but is sometimes confusing. The advantage here is that timing can be seen clearly across systems, which can be powerful for debugging 2) `npm run dev-tab` - this launches each of the required dev processes in it's own terminal session (this requires mac os grant permission to node to launch terminal sessions) - the benefit here is just having a unified way to spawn all required processes without forgetting one
- [x] Introduce some basic styles that somewhat mimic the design direction I currently have in mind (this is still WIP, but some iteration in this PR, see screenshot)
- [x] Remove `.air.toml` because we don't need Go's `air` package for this, the node scripts set up here work better since SST is a node binary
 

## UI Screenshot
image

## Concurrent (interleaved) terminal output
image

## The DaisyUI / Tailwind Theme Colors introduced in this PR

image

image


.github/workflows/deploy-prod.yml
```diff
@@ -23,6 +23,7 @@ jobs:
         with:
           MEETNEARME_TEST_SECRET: ${{ secrets.MEETNEARME_TEST_SECRET }}
           ZENROWS_API_KEY: ${{ secrets.ZENROWS_API_KEY }}
+          STATIC_BASE_URL: ${{ secrets.STATIC_BASE_URL }}
```
brianfeister, line 26: This is wrong. `sst deploy` sets this as a process.env var, but github is the parent process, while `sst deploy` is the child process. Could use suggestions here, I've got this tangled up in my head a bit 🤔

functions/lambda/views/layout.templ
```diff
@@ -5,16 +5,12 @@ import "os"
...
 		<head>
-			<link href="https://unpkg.com/prismjs@1.29.0/themes/prism-twilight.css" rel="stylesheet"/>
-			<style>
-				.token.number,
-				.token.tag {
-				  all: inherit;
-				  color: hsl(14, 58%, 55%);
-				}
-			</style>
-			<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bulma@0.9.4/css/bulma.min.css"/>
```
brianfeister, line 16: some garbage we picked up somewhere along the way, I think

stacks/StaticSiteStack.ts
```diff
@@ -0,0 +1,29 @@
...
+  const staticSite = new StaticSite(stack, 'frontend', {
+    path: 'static',
+    dev: {
+      deploy: true,
+      url: 'http://localhost:3001',
+    },
+    buildCommand: 'npm run tailwind:prod',
+    // TODO: figure out a domain name with associated cert
+    // customDomain: {
+    //   alternateNames: [api.url],
```
brianfeister, line 19: to get this working there is certificate work in Route53 that needs to happen in AWS but I haven't dug into this yet. The end result here will be `meetnear.me/assets` served from S3, while everything else serves from the Go lambda Gateway

.github/workflows/deploy-prod.yml
brianfeister: I've resolved this in this commit: https://github.com/meetnearme/api/pull/26/commits/0799b794717a34a3afd86956f8cc38a3707fbde4

package.json
```diff
@@ -7,19 +7,37 @@
     "node": ">=21.6.0"
   },
   "scripts": {
-    "dev": "sst dev",
+    "dev": "concurrently \"npm:dev:*\"",
+    "dev-tab": "concurrently \"npm:tab:*\"",
```
j4inam, line 11: TIL we could wild card npm commands :o

j4inam (approved):

package.json
brianfeister: Special power of `concurrently` :)

PR #31, opened 2024-03-22
Setup routing and middleware

j4inam:
This PR implements a web-framework to handle routing and middleware.


internal/transport/lambda/router.go
```diff
@@ -0,0 +1,129 @@
+package transport
```
brianfeister, line 1: @j4inam why is the `internal/transport/lambda` directory outside of the `functions/lambda` directory? I'm assuming both directories should contain our Go code, why not have them share a directory?

internal/transport/lambda/error_handlers.go
```diff
@@ -0,0 +1,22 @@
+package transport
+
+import (
+	"log"
+	"net/http"
+)
+
+func SendServerError(err error) (Response, error) {
```
brianfeister, line 8: In a follow-up PR, we probably want this function to have two "modes" :

1) Send an http error with a non-200 status. This would be for JSON data APIs that are leveraged by HTMX and expected to respond with HTML page partials
2) Send a HTML error page for showing an error page to the user, if for example DynamboDB fails to connect and we error out in that way

brianfeister: If that sounds agreeable, let's open a Github issue to track that

functions/lambda/handlers/page-handlers.go
```diff
@@ -0,0 +1,43 @@
...
+	"net/http"
+
+	"github.com/meetnearme/api/functions/lambda/templates/components"
+	"github.com/meetnearme/api/functions/lambda/templates/pages"
+	transport "github.com/meetnearme/api/internal/transport/lambda"
+)
+
+func GetHomePage(ctx context.Context, r transport.Request) (transport.Response, error) {
+	demoCard := components.DemoCard()
+	layoutTemplate := pages.App("Meet Near Me - Home", demoCard)
```
brianfeister, line 15: For SEO, it might be good to have all pages include `Meet Near Me` with only the `Home` as an argument passed in by each helper function. This should ultimately concatenate to `Meet Near Me - Home` with only `Home` being the string passed here

functions/lambda/main.go
```diff
@@ -1,149 +1,25 @@
...
-var validate *validator.Validate = validator.New()
+var router *transport.Router
 
-type CreateEvent struct {
-    Name string `json:"name" validate:"required"`
-    Description string  `json:"description" validate:"required"`
-    Datetime string  `json:"datetime" validate:"required"`
-    Address string  `json:"address" validate:"required"`
-    ZipCode string  `json:"zip_code" validate:"required"`
-    Country string  `json:"country" validate:"required"`
```
brianfeister, line 27: This code was being used by our `/ POST` endpoint which we should maintain for functionality. It's fine to re-structure, but I just want to confirm that we're not **removing** this `/ POST` endpoint, because we should be including it

functions/lambda/main.go
```diff
@@ -1,149 +1,25 @@
...
-		Slug:     "my-account",
-		// // Handlers: csrf.Handlers,
-	},
-	{
-		Name:     "My Events (test from non-main branch PR)",
-		Desc:     "List of events I'm attending",
-		Slug:     "my-events",
-		// // Handlers: clicktoedit.Handlers,
-	},
-}
```
brianfeister, line 44: Thought on how we want to handle a single struct that represents the overall page structure of the app. We should have one. This bit is ok to delete here, but we will want nav to be factored into our express-like framework, even if it's not derived from the declarative route style. But maybe it could be?

functions/lambda/main.go
```diff
@@ -1,149 +1,25 @@
...
-        return serverError(err)
-    }
-    // TODO: delete this?
-    log.Printf("Successfully fetched events: %s", json)
-
-    return events.APIGatewayV2HTTPResponse{
-        StatusCode: http.StatusOK,
-        Body: string(json),
-    }, nil
-}
```
brianfeister, line 94: This functionality needs to migrate to the new framework you're working on, not be deleted

functions/lambda/main.go
```diff
@@ -1,149 +1,25 @@
...
-
-func processPost(ctx context.Context, req events.APIGatewayV2HTTPRequest) (events.APIGatewayV2HTTPResponse, error) {
```
brianfeister, line 96: Mentioned above, this is our `/ POST` API call which is our only way to propagate the DB and should not be removed in the process of your migration, but rather included as a new API route + handler

functions/lambda/templates/components/demo_card.templ
```diff
@@ -0,0 +1,14 @@
+package components
+
+templ DemoCard() {
+	<div class="card w-96 bg-base-100 shadow-xl">
+		<figure><img src="https://daisyui.com/images/stock/photo-1606107557195-0e29a4b5b4aa.jpg" alt="Shoes"/></figure>
+		<div class="card-body">
+			<h2 class="card-title">Shoes!</h2>
+			<p>If a dog chews shoes whose shoes does he choose?</p>
```
brianfeister, line 8: 🤣

functions/lambda/templates/components/demo_card.templ
```diff
@@ -0,0 +1,14 @@
+package components
+
+templ DemoCard() {
```
brianfeister, line 3: Good for WIP demo code, but let's drop this one as we're not using the "card" component just yet in the final design

functions/lambda/templates/components/navbar.templ
```diff
@@ -0,0 +1,20 @@
+package components
+
+templ Navbar() {
+	<div class="navbar bg-primary shadow-xl rounded-box">
```
brianfeister, line 4: Let's preserve the styles from the navbar in `main` currently

functions/lambda/templates/pages/app.templ
```diff
@@ -0,0 +1,39 @@
+package pages
```
brianfeister, line 1: We've got `/templates/pages/*` and `template.views/*` but we should migrate the contents of `home.templ` into `app.templ` and probably `app.templ` should be called `templates/pages/home.templ` for clarity

tailwind.config.js
```diff
@@ -1,7 +1,11 @@
 /** @type {import('tailwindcss').Config} */
 
 export default {
-  content: ['functions/lambda/views/**/*.templ'],
+  content: [
+    'functions/lambda/views/**/*.templ',
+    'functions/lambda/templates/pages/**/*.templ',
+    'functions/lambda/templates/components/**/*.templ',
+  ],
```
brianfeister, line 8: I think we want a singular top-level directly like `functions/lambda/templates` so that we're not confused about where the views live. Don't want them spread out in multiple places

stacks/StaticSiteStack.ts
```diff
@@ -1,4 +1,5 @@
-import { Api, StaticSite, use } from 'sst/constructs';
+import { Api, StackContext, StaticSite, use } from 'sst/constructs';
```
brianfeister, line 1: Are we using `StackContext`? Looks like maybe not

internal/transport/lambda/router.go
j4inam: Moved transport layer within `functions/lambda`

internal/transport/lambda/error_handlers.go
j4inam: Created issue #33

functions/lambda/handlers/page-handlers.go
j4inam: Added `Meet Near Me` as hardcoded text in app layout

functions/lambda/main.go
j4inam: I moved `POST /` to `POST /events` for consistency using the new router.
We can still add events using that endpoint. The handler code is identical to the original setup.

functions/lambda/main.go
j4inam: I migrated that bit to the new handler and created a new route `GET /` which renders the homepage with events list.
This code lives in the services now: https://github.com/meetnearme/api/pull/31/files#diff-5ace55baead0c68bae661e54e511459734fff6353b55fdb96728575772d1bcc1R35

functions/lambda/main.go
j4inam: We can have declarative routing.
Need to iterate over a list of routes to build the router and then pass the same array to NavBar

functions/lambda/templates/components/demo_card.templ
j4inam: lol this is an actual card example on DaisyUI docs 😂

stacks/StaticSiteStack.ts
j4inam: Removed unused imports

scripts/sam_cli_entrypoint.sh
```diff
@@ -1,2 +1 @@
-GOOS=linux go build -o main
```
j4inam, line 1: This command explicitly tries to build go code instead of letting sst handle it. 
This bit seems to be pointing to incorrect dir as base resulting in go.mod deleting most required dependency as it assumes they're all unused which is incorrect.
Dropping this command fixes running `npm run dev` with sst and dynamodb and not breaking go build

Brandon-G-Tripp: Perfect. We probably can eliminate most of that file and comment out or remove the service in the docker compose file.

Brandon-G-Tripp (approved):

functions/lambda/handlers/data-handers.go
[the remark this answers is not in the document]
brianfeister: +1 let's fix the file name spelling error here

functions/lambda/handlers/data-handers.go
```diff
@@ -0,0 +1,62 @@
...
+
+	err = validate.Struct(&createEvent)
+
+	// TODO: Update errors to send htmx template with error message
+	if err != nil {
+		log.Printf("Invalid body: %v", err)
+		return transport.SendClientError(http.StatusBadRequest, "Invalid Body")
+	}
+
+	res, err := services.InsertEvent(ctx, db, createEvent)
```
brianfeister, line 37: Should this be `*db` to pass a pointer instead of copying `db`?

functions/lambda/main.go
```diff
@@ -1,149 +1,34 @@
...
-    json, err := json.Marshal(eventList)
-    if err != nil {
-        return serverError(err)
-    }
-    // TODO: delete this?
-    log.Printf("Successfully fetched events: %s", json)
+func init() {
+	db = transport.CreateDbClient()
+	router = transport.NewRouter()
+	router.GET("/", handlers.GetHomePage, transport.LogRequest)
```
brianfeister, line 22: This one could be an AWS cost-to-serve risk. Are we logging **ALL** GET requests? I'm pretty sure we could crush ourselves with logs / run a high bill if we log all success states. Let's only log error states, and even then, I think we don't want to log ALL error states, just the ones that are most interesting / useful

functions/lambda/transport/router.go
```diff
@@ -0,0 +1,137 @@
...
+	middleware   []Middleware
+}
+
+type Router struct {
+	routes []*route
+}
+
+func NewRouter() *Router {
+	return &Router{routes: []*route{}}
+}
```
brianfeister, line 34: Cleanup needed here, there should only be a single `Router` and not a `NewRouter` not sure which is the keeper

functions/lambda/transport/router.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (r *Router) addRoute(method string, pathDef string, handler lambdaHandlerFunc, middleware ...Middleware) *route {
+	// handle path parameters
+	pathParamMatcher := regexp.MustCompile(":([a-zA-Z]+)")
```
brianfeister, line 38: is `:` character in regex literal here? I assume this is for `/path/:varname`? If my guess is right, can you please add a dummy route that just shows how this works?

For example, add

```
router.GET("/event/:eventId", handlers.GetSingleEventPage, transport.LogRequest)
```

And then add an empty `templ` template that just shows

```
<div>Event ID: {eventId}</div>
```

We want to have a working example of this if I'm understanding correctly. Ultimately we want a test in code, but for now, let's have an example in code

functions/lambda/transport/router.go
```diff
@@ -0,0 +1,137 @@
...
+			paramKeys = append(paramKeys, matches[i][1])
+		}
+	}
+	// check for duplicates: same method and regex pattern
+	regex := regexp.MustCompile("^" + pattern + "$")
+	for _, route := range r.routes {
+		if route.method == method && route.pattern.String() == regex.String() {
+			panic(fmt.Sprintf("Route already exists: %s %s", method, pathDef))
+		}
+	}
```
brianfeister, line 56: Similar to https://github.com/meetnearme/api/pull/31/files#r1547081136 above, are we capturing the functional nuance in working project code? If not, let's add some working code that exercises this (and tests very soon!) to validate that it works as expected (and the test to ensure it continues to do so!)

functions/lambda/main.go
brianfeister: Let's not worry about this for now. Let's consider it a follow up. I actually don't think I want to expose all of the routes in the navbar ever anyway. But it could be helpful to have a helper object representation of the site to pass around, even if we don't iterate over the whole thing and stuff it in the navigation. But again, let's hold off on this and count it as a future improvement so that this PR doesn't get mired in "just one more thing-ism"

stacks/StaticSiteStack.ts
```diff
@@ -1,5 +1,4 @@
-import { Api, StaticSite, use } from 'sst/constructs';
-import { ApiStack } from './ApiStack';
+import { StackContext, StaticSite } from 'sst/constructs';
```
brianfeister, line 1: Thanks!

brianfeister (changes requested):
This looks awesome! Very excited to land this! ❗ 

A few of my comments need discussion or changes, so I'll mark as blocking for now, but I think we're very close 🚀

functions/lambda/main.go
j4inam: Opened a new issue #36 for this

functions/lambda/handlers/data-handers.go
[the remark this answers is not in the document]
j4inam: @Brandon-G-Tripp Correct. Most of the code from the original setup will live in the handlers.

functions/lambda/handlers/data-handers.go
j4inam: `db` is of type `*dynamodb.Client` so `db` is essentially a pointer to the client.

functions/lambda/main.go
j4inam: Oh that was just an example to showcase how middlewares can be used with our home grown routing. I can take that logger out.

functions/lambda/transport/router.go
j4inam: `Router` is the struct which contains the `routes` array which is the keeper or the object that actually holds the routes. Since the name `Router` is now taken, I created a `NewRouter` function that returns a pointer to the `Router` instance. I renamed it to `GetRouter` for better clarity

functions/lambda/transport/router.go
j4inam: Added `/events/:eventId` route to load the event details page.
To Test: Click on the details button in the events table on the Home page

functions/lambda/transport/router.go
j4inam: Replacing path params with values and storing passed values for params in an object is working. I've added a demo route: `/events/:eventId` to demonstrate the same.

The check for dupes bit needs a unit test to be added. Will add that once we have a unit testing framework setup.

brianfeister (changes requested):
@j4inam Definitely want to get this merged, but the deployment (we need integration tests I think) is throwing `Internal Server Error`

I checked this by going to your latest deployment here
https://github.com/meetnearme/api/actions/runs/8626305835/job/23644293424#step:5:75

And grabbing that URL and checking:
https://baqfmnplod.execute-api.us-east-1.amazonaws.com/

I'm going to pull your branch and see if I can quickly spot the problem and fix

brianfeister:
The code here looks fine in my local deployment, I'll delete `feature-frontend-setup` in AWS CloudFormation and trigger a re-deploy from fresh

functions/lambda/transport/router.go
brianfeister: Sounds good, we'll figure out something re: the dupes logic, we can merge now. 

Created the ticket here:
https://github.com/meetnearme/api/issues/37

brianfeister (approved):
LGTM 🚀

PR #38, opened 2024-04-16
Tweak / refine daisy ui theming, brand colors, add component for nested checkbox dropdown

brianfeister:
This PR introduces mostly UI changes:

- [x] - Dropdown with nested checkboxes (for event category filtering) 
Screen_Recording_2024-04-16_at_3 13 12 PM
- [x] - New daisy UI theme colors
image
image


static/assets/global.css
```diff
@@ -24,3 +24,44 @@
...
+.dropdown-content.menu .label {
+  justify-content: flex-start;
+}
+
+.dropdown-content .collapse-title,
+:where(.dropdown-content .collapse > input[type='checkbox']),
+:where(.dropdown-content .collapse > input[type='radio']) {
+  /* min-height: 0; */
+  min-height: 1.5rem;
+}
```
brianfeister, line 67: There might be a way to override some of this at the theming level in daisy UI's config, but for now, it also kinda feels good to just get surgical with CSS, so I'm ok with it for now, we can remove or alter later, if we get more sophisticated

tailwind.config.js
```diff
@@ -90,20 +90,32 @@ export default {
...
+          'neutral-content': '#00ceff',
+          'base-100': '#fffbe6',
+          info: '#7cbbee',
+          success: '#74ea62',
+          warning: '#ffc458',
+          error: '#ff7f7f',
+          primary: '#00ceff',
+          secondary: '#5eead4',
+          accent: '#f0abfc',
+          neutral: '#190c04',
```
brianfeister, line 118: Surfacing **all** of the underlying theme's attributes for specific override for declarative control

brianfeister: (many of these come from `cyberpunk`, but using `cupcake` theme because of some deep stuff in `cyberpunk` I couldn't figure out how to override)

functions/lambda/helpers/constants.go
```diff
@@ -5,3 +5,338 @@ const EventsTablePrefix = "Events"
...
+type Category struct {
+	Name, Desc, Slug string
+	Items []Subcategory
+}
+
+type Subcategory struct {
+	Name, Desc, Slug string
+}
+
+var Categories = []Category {
```
brianfeister, line 18: This is WIP, obviously, but it's good to get these types in place now and iterate when we are ready to connect them to the DB for things like `slug`, `uuid` and others. Depends on our architecture for this on the backend, so leaving as placeholder for now

static/assets/global.css
Brandon-G-Tripp: Great to have that fine tuning ability with daisy still.

Brandon-G-Tripp (approved):
This is looking really good to me. I am excited to get that matched up with some of this copy writing and have it throughout the UI. 

Good placeholder stuff for the categories. Good to keep in mind. Think we can iterate on the schema to decide how to categorize and tag. That will be a more challenging aspect of the schema I imagine. 

Looks good to me.

PR #39, opened 2024-05-09
feature/z order index creation and query

Brandon-G-Tripp:
Integrate Z-Order Indexing for Efficient Event Retrieval

Description:
This pull request integrates the Z-order indexing functionality into the event retrieval process. Z-order indexing allows for efficient querying of events based on location (latitude and longitude) and time range. By utilizing this indexing technique, we can significantly improve the performance of event retrieval and provide a better user experience.

The following changes have been made:

1. Implemented the Z-order indexing logic in the `indexing` package.
2. Updated the `Event` struct to include fields for latitude, longitude, and the Z-order index.
3. Modified the `GetEvents` function in the `events_service` to handle start and end time in the `QueryInput`.
4. Updated the `InsertEvent` function in the `events_service` to calculate the Z-order index based on latitude, longitude, and time.
5. Added latitude and longitude fields to the `EventSelect` and `EventInsert` structs.
6. Updated the `GetHome` handler in `page-handlers.go` to use the Z-order indexing for event retrieval.
7. Added examples of coordinates for inserting new events in different regions and times.

To test the Z-order indexing functionality, you can use the following curl commands to insert new events:

```
curl -X POST -H "Content-Type: application/json" -d '{
  "name": "New York Event",
  "description": "An event in New York",
  "datetime": "2023-06-01T10:00:00Z",
  "address": "123 Main Street",
  "zip_code": "10001",
  "country": "USA",
  "latitude": 40.7128,
  "longitude": -74.0060
}' https://<your-deployment-url>/api/event

curl -X POST -H "Content-Type: application/json" -d '{
  "name": "London Event",
  "description": "An event in London",
  "datetime": "2023-07-15T14:30:00Z",
  "address": "456 Oxford Street",
  "zip_code": "W1A 1AB",
  "country": "UK",
  "latitude": 51.5074,
  "longitude": -0.1278
}' https://<your-deployment-url>/api/event

curl -X POST -H "Content-Type: application/json" -d '{
  "name": "Tokyo Event",
  "description": "An event in Tokyo",
  "datetime": "2023-08-10T09:00:00Z",
  "address": "789 Shibuya Street",
  "zip_code": "150-0002",
  "country": "Japan",
  "latitude": 35.6895,
  "longitude": 139.6917
}' https://<your-deployment-url>/api/event
```

Make sure to replace `<your-deployment-url>` with the actual URL of your deployed API.

To retrieve each event individually using the `GetHome` handler and demonstrate that it does not query the others, you can use the following hardcoded parameters:

1. New York Event:
```go
startTime := time.Date(2023, 6, 1, 9, 0, 0, 0, time.UTC)
endTime := time.Date(2023, 6, 1, 11, 0, 0, 0, time.UTC)
lat := float32(40.7128)
lon := float32(-74.0060)
radius := float32(100.0)
```

2. London Event:
```go
startTime := time.Date(2023, 7, 15, 13, 0, 0, 0, time.UTC)
endTime := time.Date(2023, 7, 15, 15, 0, 0, 0, time.UTC)
lat := float32(51.5074)
lon := float32(-0.1278)
radius := float32(100.0)
```

3. Tokyo Event:
```go
startTime := time.Date(2023, 8, 10, 8, 0, 0, 0, time.UTC)
endTime := time.Date(2023, 8, 10, 10, 0, 0, 0, time.UTC)
lat := float32(35.6895)
lon := float32(139.6917)
radius := float32(100.0)
```

By using these specific parameters for each event, you can demonstrate that the Z-order indexing correctly retrieves only the desired event without querying the others. The narrow time range and small radius ensure that only the target event is returned.

Please note that the Z-order indexing implementation assumes a maximum precision of 1e-6 for latitude and longitude values. Events with coordinates beyond this precision may not be indexed accurately.

Commits:
- **Create JSON schema for the event model as source of truth in the SST stacks and in Go API**
- **Move of the event_schema.json file to internal/models dir**
- **Go Event struct to model data for JS and dynamodb**
- **Update Event struct field for ZOrderIndex to []byte**
- **Implementation of the ZOrderIndexing and the mapFloatToSortableInt functions in the indexing package**
- **Add z order index into the event select and event insert structs along with lat and lon fields.**
- **Add fields to storage stack for lat, lon and z index.**
- **Update the GetEvents function to handle start and end time in the QueryInput**
- **feat(z_order_index): Add latitude and longitude to the event service**
- **Add dev-remote-db script command in package.json**
- **Update date format helpers to use time.RFC3339**
- **feat(z_order_indexing): Add indexing testing and logic for z order index**
- **feat(z_order_index): InsertEvent in events_services to calculatre z order index**
- **feat(z_order_indexing): change lat/long to 32 bits on EventSelect and EventInsert**
- **feat(z_order_indexing): func to get events by z order range (lat, long, start time and appended creation time)**
- **feat(z_order_indexing): z order index calculation and index creation**
- **feat(z_order_indexing): adding the z order indexing to HomePage handler**
- **Removal of logging in events_service.go**


functions/lambda/handlers/page-handlers.go
```diff
@@ -14,19 +16,55 @@ import (
...
+    latStr := r.QueryStringParameters["lat"]
+    lonStr := r.QueryStringParameters["lon"]
+    radiusStr := r.QueryStringParameters["radius"]
+
+    // Set default values if query parameters are not provided
+    startTime := time.Now()
+    endTime := startTime.AddDate(100, 0, 0)
+    lat := float32(39.8283)
+    lon := float32(-98.5795)
+    radius := float32(2500.0)
```
brianfeister, line 31: This is probably must me not understanding golang... but, shouldn't these be `=` as opposed to `:=` since they've already been declared above? And don't you need `if <condition> {}` around these to only override in the case of the values still being `nil` at these lines?

functions/lambda/services/events_service.go
```diff
@@ -19,6 +23,9 @@ type EventSelect struct {
 	Address     string `json:"address" dynamodbav:"address"`
 	ZipCode     string `json:"zip_code" dynamodbav:"zip_code"`
 	Country     string `json:"country" dynamodbav:"country"`
+    Latitude float32 `json:"latitude" dynamodbav:"latitude"`
+    Longitude float32 `json:"longitude" dynamodbav:"longitude"`
+    ZOrderIndex []byte `json:"z_order_index" dynamodbav:"zOrderIndex,B"`
```
brianfeister, line 28: lol, golang indentation

brianfeister (commented):
@Brandon-G-Tripp Quick question here – I'm very interested in the problem of "how can we avoid a bunch of cruft around inconsistent event data with various schemas over time?" Granted, we're NoSQL, so we don't have big db migrations, but we **do** still need to try to get this right as early as possible so we're not constantly tweaking and shooting ourselves in the foot

More specific (related) question – Is it safe for me to pull this branch locally and test it without blowing up my database? I want us to be thinking about backwards compatibility 

But I also recognize that the initial "schema" definition for the data we have in our DB currently is woefully bad and **should not** be supported.

Thoughts?

functions/lambda/indexing/z_order_index.go
```diff
@@ -0,0 +1,86 @@
+package indexing
+
+import (
+	"encoding/binary"
+	// "encoding/hex"
+	"fmt"
+	"math"
+	"strconv"
+	// "strings"
```
brianfeister, line 9: remove

functions/lambda/indexing/z_order_index.go
```diff
@@ -0,0 +1,86 @@
+package indexing
+
+import (
+	"encoding/binary"
+	// "encoding/hex"
```
brianfeister, line 5: remove

functions/lambda/indexing/z_order_index.go
```diff
@@ -0,0 +1,86 @@
...
+    startTimeBin := strconv.FormatInt(startTimeUnix, 2)
+    startTimeBin = fmt.Sprintf("%032s", startTimeBin)
+
+    // Map floating point values to sortable unsigned integers
+    lonSortableInt := mapFloatToSortableInt(lon)
+    latSortableInt := mapFloatToSortableInt(lat)
+
+    // Convert sortable integers to binary string
+    lonBin := strconv.FormatUint(uint64(lonSortableInt), 2)
+    lonBin = fmt.Sprintf("%032s", lonBin)
```
brianfeister, line 28: This line is padding the binary to a length of `32`?

functions/lambda/services/events_service.go
```diff
@@ -63,26 +73,92 @@ func GetEvents(ctx context.Context, db *dynamodb.Client) ([]EventSelect, error)
...
+
 	newEvent := EventSelect{
 		Name:        createEvent.Name,
 		Description: createEvent.Description,
 		Datetime:    createEvent.Datetime,
 		Address:     createEvent.Address,
 		ZipCode:     createEvent.ZipCode,
+        Latitude: createEvent.Latitude,
+        Longitude: createEvent.Longitude,
+        ZOrderIndex: zOrderIndex,
```
brianfeister, line 140: Would be nice if there were some linter / auto-fix for this weird golang indentation thing

functions/lambda/services/events_service.go
```diff
@@ -63,26 +73,92 @@ func GetEvents(ctx context.Context, db *dynamodb.Client) ([]EventSelect, error)
...
 
+    log.Printf("%v", newEvent)
+
 	item, err := attributevalue.MarshalMap(newEvent)
 	if err != nil {
+        log.Print("Error in marshall!!!!")
 		return nil, err
 	}
 
+    log.Printf("Item before insert %v", item)
```
brianfeister, line 152: Cleanup

functions/lambda/services/events_service.go
```diff
@@ -63,26 +73,92 @@ func GetEvents(ctx context.Context, db *dynamodb.Client) ([]EventSelect, error)
...
 		Datetime:    createEvent.Datetime,
 		Address:     createEvent.Address,
 		ZipCode:     createEvent.ZipCode,
+        Latitude: createEvent.Latitude,
+        Longitude: createEvent.Longitude,
+        ZOrderIndex: zOrderIndex,
 		Id:          uuid.NewString(),
 	}
 
+    log.Printf("%v", newEvent)
```
brianfeister, line 144: Cleanup

functions/lambda/services/events_service.go
```diff
@@ -63,26 +73,92 @@ func GetEvents(ctx context.Context, db *dynamodb.Client) ([]EventSelect, error)
...
 	}
 
+    log.Printf("Item before insert %v", item)
+
 	input := &dynamodb.PutItemInput{
 		TableName: aws.String(TableName),
 		Item:      item,
 	}
 
+    log.Printf("Item DB input  %v", input)
```
brianfeister, line 159: Cleanup

internal/models/event_schema.json
```diff
@@ -0,0 +1,45 @@
+{
```
brianfeister, line 1: Nice, glad to see this in JSON schema? Are we actually using it though or is it just an artifact of attempting to?

package.json
```diff
@@ -11,6 +11,7 @@
     "dev-tab": "concurrently \"npm:tab:*\"",
     "build": "sst build",
     "dev:dynamodb": "docker compose up",
+    "dev-remote-db": "concurrently \"npm:dev:*(!dynamodb)\"",
```
brianfeister, line 14: 👍

brianfeister (commented):
@Brandon-G-Tripp great PR description btw! But can you add an explanation of the need to use the `dev-remote-db` command for our future selves (and @j4inam) when validating this locally on our side?

We can't do this with local docker / sam dynamo right?

brianfeister:
@Brandon-G-Tripp finally attempting to check this out and test. Documenting the results of that attempt here:

1. First, I switched and hit this error `CloudFormation cannot update a stack when a custom-named resource requires replacing. Rename meetnearme-api-stage-bfeister-meetnearme-go-fullstack-Events and update the stack again.`

This seems to confirm my suspicion that I'll need to delete my Dynamo table because SST doesn't do any automatic amendment when the schema changes (in this case, you've changed it by adding the `partitionKey` and new field.

image

brianfeister:
@Brandon-G-Tripp Working for me locally now! 🔥 

Super exciting! 🚀  ... I'm tinkering with getting PR checks in Github Actions CI working with this PR

image

image

image

functions/lambda/handlers/page-handlers.go
Brandon-G-Tripp: We are setting default values for the query parameters that we need to scan with, then lower we check if those are present on the ctx struct that is passed in. If they are not present we just use our set default range of time and lat/long.

internal/models/event_schema.json
Brandon-G-Tripp: This is the first draft of schema and is an artifact at this point. Will remove and reintroduce in future PR for the JSON schema.

brianfeister (approved):
Let's get this merged 🚀

PR #40, opened 2024-06-07
Integrate clerk auth

j4inam:
## Description
This PR implements authentication via Clerk using clerk-go-sdk v2.

## Changes
- Implement clerk middleware
- Implement login and signup pages
- Implement error handling response for HTTPErrors and displaying fallback htmx error components.\

## How to test this PR
- Checkout the branch
- Run `npm run dev`
- Navigate to Home page
- Click on **Account** in the navbar. (You should be redirected to the login page)
- Click on signup and sign up using Email
- Post successful signup, Login using Email from the login screen
- Click on **Account** in the navbar again. You should see your account name and userID on screen. Refer screenshots.

## Screenshots
Screenshot 2024-06-07 at 4 53 27 PM

Screenshot 2024-06-07 at 4 53 41 PM

Screenshot 2024-06-07 at 4 54 13 PM


functions/lambda/handlers/data-handlers.go
```diff
@@ -16,36 +16,53 @@ import (
...
 	// TODO: Update errors to send htmx template with error message
 	if err != nil {
 		log.Printf("Invalid JSON payload: %v", err)
-		return transport.SendClientError(http.StatusUnprocessableEntity, "Invalid JSON payload")
+		return transport.SendHTTPError(&transport.HTTPError{
+			Status:          http.StatusUnprocessableEntity,
+			Message:         "Invalid JSON payload",
+			ErrorComponent:  nil,
+			ResponseHeaders: map[string]string{"Content-Type": "application/json"},
+		})
```
brianfeister, line 31: Nice 👍

brianfeister (commented):
@j4inam can you add the open TODO's on this or file them as bugs / tickets? The only one I remember is my comment about the password input field being required for launch, I don't want us to do "magic links" only for sign in.

You had mentioned 3 other things, I believe?

I would like to get this merged so we can all have this in `main` to integrate with our own feature work, and I think I'd like to merge it before those changes are in place, I just want to make sure they are tracked

functions/lambda/handlers/page-handlers.go
```diff
@@ -63,12 +96,49 @@ func GetEventDetailsPage(ctx context.Context, r transport.Request, db *dynamodb.
...
 		Headers:         map[string]string{"Content-Type": "text/html"},
 		StatusCode:      http.StatusOK,
 		IsBase64Encoded: false,
 		Body:            buf.String(),
-	}, nil
+	}
+}
+
+func GetAccountPage(ctx context.Context, r transport.Request, db *dynamodb.Client, clerkAuth *transport.ClerkAuth) transport.Response {
+	sessionClaims, ok := clerk.SessionClaimsFromContext(ctx)
```
brianfeister, line 114: Not a blocker, but we should demonstrate other session claims (e.g. "roles") as a follow-up. Not required for this PR though

functions/lambda/helpers/constants.go
```diff
@@ -1,7 +1,8 @@
 package helpers
 
-const EventsTablePrefix = "Events"
-
 const (
-	EVENT_ID_KEY string = "eventId"
+	EventsTablePrefix        = "Events"
+	EVENT_ID_KEY      string = "eventId"
+	GOOGLE_AUTH       string = "oauth_google"
```
brianfeister, line 6: I saw Discord as an auth option when we looked at your WIP project via video call... does this need to include something regarding Discord? Also, I'm not sure that we really need Discord auth, seems a little less common, and also, I feel like we might want Outlook (Microsoft) auth, since we are calendar-focused and want to add events to calendars as RSVPs

functions/lambda/main.go
```diff
@@ -2,28 +2,52 @@ package main
...
+
+	clerk.SetKey(apiKey)
+
+	clerkAuth = transport.InitClerkAuth(config)
+
+	// Setup Routing
 	router = transport.GetRouter()
 	router.GET("/", handlers.GetHomePage)
 	router.GET("/login", handlers.GetLoginPage)
+	router.GET("/sign-up", handlers.GetSignUpPage)
```
brianfeister, line 41: Minor nit, let's do `signup` I like to avoid dashes unless a word REALLY needs to be clarified by a delimiter

functions/lambda/transport/auth.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+type ClerkAuth struct {
+	GenericClient *client.Client
+	UserClient    *user.Client
+	SessionClient *session.Client
+}
+
+func InitClerkAuth(config *clerk.ClientConfig) *ClerkAuth {
+	genericClient := client.NewClient(config)
```
brianfeister, line 17: What is `genericClient`? I feel like we might want a better name for this, as it's not immediately obvious to me what this refers to

functions/lambda/transport/error_handlers.go
```diff
@@ -1,22 +1,45 @@
...
+type HTTPError struct {
+	Status          int
+	Message         string
+	ErrorComponent  templ.Component
+	ResponseHeaders map[string]string
+}
 
+// Writes message and status of HTTPError to the response
+// Returns false if httpError is nil, true otherwise
+func SendHTTPError(httpError *HTTPError) Response {
```
brianfeister, line 20: Awesome 👏

functions/lambda/transport/middleware.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+	"strings"
+
+	"github.com/clerk/clerk-sdk-go/v2"
+	"github.com/clerk/clerk-sdk-go/v2/http"
+	"github.com/clerk/clerk-sdk-go/v2/jwt"
+	"github.com/meetnearme/api/functions/lambda/helpers"
+)
+
+func ParseCookies(ctx context.Context, r Request) (context.Context, Request, *HTTPError) {
```
brianfeister, line 13: Interesting ... there's no cookie parser function baked into Go's core libs?

functions/lambda/templates/pages/signup.templ
```diff
@@ -0,0 +1,18 @@
+package pages
+
+templ SignUpPage() {
+	<script async crossorigin="anonymous" data-clerk-publishable-key="pk_test_YnJhdmUtYmVldGxlLTEyLmNsZXJrLmFjY291bnRzLmRldiQ" src="https://brave-beetle-12.clerk.accounts.dev/npm/@clerk/clerk-js@latest/dist/clerk.browser.js" type="text/javascript"></script>
```
brianfeister, line 4: Is this a secret? Probably should come from env var, right?

functions/lambda/transport/middleware.go
```diff
@@ -0,0 +1,64 @@
...
+
+func RequireHeaderAuthorization(ctx context.Context, r Request) (context.Context, Request, *HTTPError) {
+	cookiesMap := ctx.Value("cookiesMap").(map[string]string)
+	sessionToken := cookiesMap[helpers.SESSION_COOKIE]
+	if sessionToken == "" {
+		httpError := &HTTPError{
+			Status:          302,
+			Message:         "Unauthorized. Session token missing.",
+			ErrorComponent:  nil,
+			ResponseHeaders: map[string]string{"Location": "https://zi01imi18j.execute-api.us-east-1.amazonaws.com/login?redirect=" + r.RawPath},
```
brianfeister, line 34: This is a blocker ❌ 

Needs to use `os.Getenv('API_APEX_URL')` or something similar

Needs to be emitted to `.env` as part of the SST env vars that get exposed here:

https://github.com/meetnearme/api/blob/feature-clerk-auth/stacks/ApiStack.ts#L22

Notice that we emit it to console here 
https://github.com/meetnearme/api/blob/feature-clerk-auth/stacks/ApiStack.ts#L34

That prints it for us to know when we run `npm run dev` locally`, but we need to add it to the above section for env vars so that it's exposed to the Go runtime and can be dynamically set here

functions/lambda/templates/pages/login.templ
```diff
@@ -1,7 +1,36 @@
 package pages
 
 templ LoginPage() {
-	<section class="w-full">
-		<h1 class="text-2xl">Login Page</h1>
+	<script async crossorigin="anonymous" data-clerk-publishable-key="pk_test_YnJhdmUtYmVldGxlLTEyLmNsZXJrLmFjY291bnRzLmRldiQ" src="https://brave-beetle-12.clerk.accounts.dev/npm/@clerk/clerk-js@latest/dist/clerk.browser.js" type="text/javascript"></script>
```
brianfeister, line 4: Same question as here:

https://github.com/meetnearme/api/pull/40/files#r1633937284

brianfeister (changes requested):
This one especially is a blocker:
https://github.com/meetnearme/api/pull/40/files#r1633939331

A few others are "maybe a blocker" and then some of the nits are small enough that we should probably just fix them now.

Let me know your thoughts, really want to get this merged, and please add to the PR description of open / unresolved issues for tracking. We can merge with those feature defects

j4inam:
> @j4inam can you add the open TODO's on this or file them as bugs / tickets? The only one I remember is my comment about the password input field being required for launch, I don't want us to do "magic links" only for sign in.
> 
> You had mentioned 3 other things, I believe?
> 
> I would like to get this merged so we can all have this in `main` to integrate with our own feature work, and I think I'd like to merge it before those changes are in place, I just want to make sure they are tracked

Opened issues #42 #43 #44

functions/lambda/helpers/constants.go
j4inam: I added it as a login option just to demo clerk we can remove discord and keep only email (link + password) and Google login. I disabled Discord login via dashboard.

functions/lambda/main.go
j4inam: Fixed! Refactored to `/signup`

functions/lambda/transport/middleware.go
j4inam: The built in function is for accessing cookies from Go's net/http request. Doesn't really work for Lambdas.
The object structures are different

functions/lambda/transport/middleware.go
j4inam: I created an issue #45 to address this

functions/lambda/transport/auth.go
j4inam: Generic Client is for accessing global clerk features that don't really fall under a specific category like user or organization or session, etc. I can rename it to `clerkClient` or `globalClient` if that adds clarity

functions/lambda/templates/pages/login.templ
j4inam: This is a publishable secret specifically for being used on the front end and it doesn't matter if the value is exposed.
I can move it to env var for code quality purposes but other than that having the exposed is fine.

There's also a server side secret which should not be exposed.

See: https://clerk.com/docs/deployments/clerk-environment-variables#clerk-publishable-and-secret-keys

functions/lambda/transport/middleware.go
brianfeister: @j4inam this will break me and @Brandon-G-Tripp if the PR merges with this AWS URL. The URL is specific to your deployment on AWS so we can't use this. I've pushed the commit below to demonstrate what I had in mind for resolving this. This should give us a way to override the URL in a way that's specific to each developer's environment

https://github.com/meetnearme/api/pull/40/commits/3aea3199a6e795794507ea0cfcdb8b030dc20807

Let me know if this change works on your side.

On my end, I can't test the PR because I think I have to delete the whole AWS resource array because I guess the Dynamo table change Brandon introduced isn't included here? Not sure 🤔 did you merge upstream `main` into this?

brianfeister:
@j4inam looks like you need to merge `main` back. Then hopefully I can get a clean deploy with the Dynamo schema change from @Brandon-G-Tripp included on your branch here

brianfeister:
@j4inam Ideally we name the branch `feature/<my-feature-name>` which triggers the feature branch deployment which would be helpful for testing here. This branch is `feature-clerk-auth` and the lack of `/` after feature fails the regex test for triggering the deploy 

https://github.com/meetnearme/api/blob/main/.github/workflows/deploy-feature.yml#L5

functions/lambda/transport/middleware.go
```diff
@@ -0,0 +1,68 @@
...
+	"strings"
+
+	"github.com/clerk/clerk-sdk-go/v2"
+	"github.com/clerk/clerk-sdk-go/v2/http"
+	"github.com/clerk/clerk-sdk-go/v2/jwt"
+	"github.com/meetnearme/api/functions/lambda/helpers"
+)
+
+func InitAppContext(ctx context.Context, r Request) (context.Context, Request, *HTTPError) {
+	domainName := r.RequestContext.DomainName
```
j4inam, line 14: Fixes #45 . Writes domain name to `context.Context` instead of .env

brianfeister (changes requested):
@j4inam I really want to get this merged. Here are the issues I'm seeing testing this on my side

1.  The first thing that happened for me running this locally (I was able to now, yay 🎉 ) was going to `/account` throws an error
image
2. When I click "Sign In" I can't get back to the login form to sign in (I see the open bug for 30 second sessions, but how can I log in at all? Do we need some javascript that will clear cookies when clicking "sign in" in order for the backend middleware to not pick that up and try to throw it through auth refresh flow? Screenshot of the issue:
![Screen_Recording_2024-06-23_at_10 17 27 PM] (https://github.com/meetnearme/api/assets/806536/5328380c-a64a-4cf9-9207-8c3e577d8c84)

functions/lambda/handlers/data-handlers.go
```diff
@@ -16,36 +16,53 @@ import (
 
 var validate *validator.Validate = validator.New()
 
-func CreateEvent(ctx context.Context, r transport.Request, db *dynamodb.Client) (transport.Response, error) {
+func CreateEvent(ctx context.Context, r transport.Request, db *dynamodb.Client, clerkAuth *transport.ClerkAuth) transport.Response {
```
brianfeister, line 19: @j4inam I think we want this available via middleware so that we're not constantly changing the function signature to something non-standard. That way, only the routes that need a given parameter need to have their code changed. Especially in a situation like this where `clerkAuth` can be a pointer so we're not doing the JavaScript mistake of passing everything by value

brianfeister:
Unless something changes, we are going with Zitadel auth instead of Clerk:
https://github.com/meetnearme/api/pull/50

PR #47, opened 2024-06-28
Migrate to gorrila/mux router, `go mod tidy` for outdated dependencies

brianfeister:
- [x] This PR migrates from our custom router solution to `gorilla/mux` because during the process of exploring authentication SDKs, I discovered most of them expect the pattern found in `net/http` which is to write to `http.ResponseWriter` (which is a streamed byte array). Rather than return a response, this moves us to a `return http.HandlerFunc(nil)` allowing the underlying `net/http` pattern of picking up the following middleware via `next()` which can further modify the response byte buffer. I believe that internally the router responds when automatically when it gets to the last item in the middleware chain.

- [x] The original AWS `events.APIGatewayV2HTTPRequest` and `events.APIGatewayV2HTTPRequestContext` are now made available via this pattern:

```
func GetHomePage(w http.ResponseWriter, r *http.Request, db *dynamodb.Client) http.HandlerFunc {
	// Extract parameter values from the request query parameters
	ctx := r.Context()
	apiGwV2Req := ctx.Value(helpers.ApiGwV2ReqKey).(events.APIGatewayV2HTTPRequest)

	queryParameters := apiGwV2Req.QueryStringParameters
	startTimeStr := queryParameters["start_time"]
         ...
}
```
- [x] This change also allows us to use the aws sdk `github.com/awslabs/aws-lambda-go-api-proxy` for mapping `events.APIGatewayV2HTTPRequest(Context)` events to something compatible with `net/http` which will likely be needed for other functionality beyond just what I ran into with auth SDKs


Brandon-G-Tripp (approved):
This all looks good to me. No comments as it mainly follows the same Go patterns which makes it easy to review. Feels cleaner and nice having the transport http file to help. Looks good.  I will checkout the discussion when I can and then also review more in depth the seshu PR with testing local once I get setup with the google maps api.

functions/lambda/main.go
```diff
@@ -2,30 +2,84 @@ package main
...
-	router.POST("/api/event", handlers.CreateEvent)
+// Middleware to inject context into the request
+func withContext(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		// Add context to request
+		r = r.WithContext(ctx)
+		next.ServeHTTP(w, r)
+	})
 }
```
j4inam, line 36: NIT: Want to pull this out into middleware.go ?

j4inam (approved):
Small NIT. LGTM 🚀

functions/lambda/main.go
brianfeister: I think in this case, since we don't have 3 middleware yet, I prefer to keep it in one place as long as we can. Also, this is "core" functionality that affects all of the flow in `main.go` so I prefer to keep it easily visible as `r.context()` is central to how we'll be passing data, so it feels less "magic" if it's all in one file

PR #48, opened 2024-06-30
Initial `Event` JSON schema draft proposal (PLEASE DISCUSS 🙏) for `event.schema.json`

brianfeister:
# How to test

Validate the input constraints by going to https://json-editor.github.io/json-editor/ and entering the JSON schema content of the proposed JSON file in this PR.

Screen_Recording_2024-06-30_at_8 42 27 AM


architecture/event.schema.json
```diff
@@ -0,0 +1,109 @@
...
+        "required": ["recurrenceId"]
+      }
+    },
+    {
+      "required": ["recurrenceId"],
+      "not": {
+        "required": ["ownerId"]
+      }
+    }
+  ]
```
brianfeister, line 108: The `required` array doesn't include either `ownerId` or `recurrenceId`, but one of the two (but not both) is required.

This reflects the fact that a non-recurring event must have a single owner -- OR -- for recurring events that are generated from the `EventRecurrences` table, there must be a `recurrenceId`

architecture/event.schema.json
```diff
@@ -0,0 +1,109 @@
...
+    },
+    "locationAddress": {
+      "type": "string"
+    },
+    "locationCountry": {
+      "type": "string"
+    },
+    "locationZipCode": {
+      "type": "string"
+    },
```
brianfeister, line 62: I flattened these, along with the `url*` keys to simplify and avoid unnecessary nesting

architecture/event.schema.json
```diff
@@ -0,0 +1,109 @@
...
+    "name",
+    "startTime",
+    "url",
+    "urlQueryParams",
+    "locationLatitude",
+    "locationLongitude",
+    "tags",
+    "categories",
+    "createdAt",
+    "updatedAt"
```
brianfeister, line 93: This is a bit indirect, but anything not in this list is not `required`. It might be good to generate TypeScript types to go along with these (or Go structs) so that when you look at a key, you don't have to go to the bottom of the file to discover whether the field you're looking at is required

architecture/event.schema.json
```diff
@@ -0,0 +1,109 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Event",
+  "type": "object",
+  "properties": {
+    "id": {
+      "type": "string",
+      "format": "uuid"
+    },
+    "ownerId": {
```
brianfeister, line 10: Optional with the `oneOf` logic caveat below

architecture/event.schema.json
```diff
@@ -0,0 +1,109 @@
...
+      "format": "date-time"
+    },
+    "endTime": {
+      "type": "string",
+      "format": "date-time"
+    },
+    "recurrenceRule": {
+      "type": "string"
+    },
+    "recurrenceId": {
```
brianfeister, line 31: Optional with the `oneOf` logic caveat below

architecture/event.schema.json
```diff
@@ -0,0 +1,109 @@
...
+      "format": "uuid"
+    },
+    "ownerId": {
+      "type": "string",
+      "format": "uuid"
+    },
+    "name": {
+      "type": "string"
+    },
+    "description": {
```
brianfeister, line 17: optional

architecture/event.schema.json
```diff
@@ -0,0 +1,109 @@
...
+    },
+    "startTime": {
+      "type": "string",
+      "format": "date-time"
+    },
+    "endTime": {
+      "type": "string",
+      "format": "date-time"
+    },
+    "recurrenceRule": {
```
brianfeister, line 28: optional

architecture/event.schema.json
```diff
@@ -0,0 +1,109 @@
...
+    },
+    "recurrenceId": {
+      "type": "string",
+      "format": "uuid"
+    },
+    "url": {
+      "type": "string",
+      "format": "uri"
+    },
+    "urlDomain": {
```
brianfeister, line 39: optional

architecture/event.schema.json
```diff
@@ -0,0 +1,108 @@
...
+      "type": "string"
+    },
+    "urlPath": {
```
brianfeister, line 42: optional

architecture/event.schema.json
```diff
@@ -0,0 +1,108 @@
...
+      "type": "string"
+    },
+    "urlQueryParams": {
```
brianfeister, line 45: optional

architecture/event.schema.json
```diff
@@ -0,0 +1,108 @@
...
+      "type": "object"
+    },
+    "locationLatitude": {
+      "type": "number"
+    },
+    "locationLongitude": {
+      "type": "number"
+    },
+    "locationAddress": {
```
brianfeister, line 54: optional

architecture/event.schema.json
```diff
@@ -0,0 +1,108 @@
...
+      "type": "string"
+    },
+    "locationCountry": {
```
brianfeister, line 57: optional

architecture/event.schema.json
```diff
@@ -0,0 +1,108 @@
...
+      "type": "string"
+    },
+    "locationZipCode": {
```
brianfeister, line 60: optional

brianfeister:
@Brandon-G-Tripp @j4inam One important bit is to consider and propose any data we might need that isn't included in the proposal already

brianfeister:
One field I can think of is `maxParticipants` or `eventCapacity` (for venue management, which we would ideally handle in the future). Another could be `eventCost` (which would be ideal to include as early as possible so that users can filter based on their budget if they're looking for something to do)

j4inam:
Moved to #62

PR #50, opened 2024-07-09
Zitadel Auth

brianfeister:
What this PR covers

- [x] Zitadel auth
- [x] implement callback route handlers (no JS 🎉 )
- [x] optional required / check / none for auth on routes
- [x] move dynamodb and zitadel auth middleware to singleton pattern to avoid unneeded pass via params
- [x] implement "profile" page
- [x] show avatar + user info in navbar if logged in
- [ ] **TODO: No login page yet** (this PR can be tested via logging in at https://meet-near-me-production-8baqim.zitadel.cloud/ui/login/login with `info@meetnear.me` and the shared password for other accounts

## Screenshot

Screen_Recording_2024-07-08_at_7 05 33 PM-min


functions/gateway/handlers/page-handlers.go
```diff
@@ -17,7 +22,29 @@ import (
 	"github.com/meetnearme/api/functions/gateway/transport"
 )
 
-func GetHomePage(w http.ResponseWriter, r *http.Request, db *dynamodb.Client) http.HandlerFunc {
+var Db *dynamodb.Client
+var mw   *authentication.Interceptor[*openid.UserInfoContext[*oidc.IDTokenClaims, *oidc.UserInfo]]
```
Brandon-G-Tripp, line 26: My one thought is if we should alias the zitadel authentication on the import so that is clear what authentication we are using. Or do you feel it is clear enough throughout the code?

Brandon-G-Tripp (commented):
Ok everything is looking good. I looked over all of the files and now am working on testing it locally.

functions/gateway/templates/pages/profile.templ
Brandon-G-Tripp, line 1: This is looking good. I missed this profile page on the first pass but everything looks nice and functional. It is definitely getting close on the code.

Brandon-G-Tripp:
@brianfeister I did not see if the logout functionality is built into the app yet? I tried out the /auth/logout path and the logout button but it did not seem to change the authentication status.

functions/gateway/handlers/data-handers.go
Brandon-G-Tripp, line 1: It looks like this file got renamed on accident and lost the l in handlers.

Brandon-G-Tripp (approved):
Everything looks pretty good to me so far. I do not see any major changes and I am assuming the logout functionality was not included on purpose.

functions/gateway/handlers/data-handers.go
brianfeister: Ha, nice catch, not sure how long it was like that, but it's funny how filenames don't matter in Go. You just export to a package and then that's the bit you reference internally rather than the filename itself (Go just compiles and jams everything with `*.go` in the filename and then assigns it to it's `package` as directed)

brianfeister: While we're at it, I'm pushing a change to migrate all the handler files that use dash `-` instead of `_` as a delimiter. Underscore seems to be the Go file naming convention

PR #51, opened 2024-07-09
Seshu Pt 2 🍴 

### LINKED ISSUE #33: [Enhancement] Split error handler in transport layer to handle data API errors as well as UI response errors
The `ServeHTTP` function in the router returns a browser 404 if a route is not found or a 405 if a method is not allowed for a given route. 

Both these errors are browser responses. 
We need to handle cases where a request is a data API request from HTMX which might result in an error state. This requires an htmx template as a response with correct error message in the UI instead of a plain browser 400 or 404

### LINKED ISSUE #42: [Bug][Auth] Implement auto-refresh of __session token while the user is logged in.
As a user, I'd want my session to be valid for 180 days. 

Current Behavior: __session token becomes invalid in 30sec and does not refresh even if the refresh_token and session is valid.

### Steps to reproduce
- Login using email
- Navigate to Account page
- Wait for ~30sec
- Refresh the page
- Notice the 403 Invalid token error

Expected behavior: __session token must automatically refresh as long as the user is logged in and the session has not expired.

### LINKED ISSUE #43: [Bug][Auth] Login with Google does not set __session cookie
As a user I want to be able to login into the app using Google OAuth. 

## Current Behavior
After successful authentication, google OAuth fails to set the session token in the __session cookie

## Steps to Reproduce
- Navigate to login page
- Click on Login with Google
- Login into your google account
- Allow app access on Google consent screens
- You'll be navigated back to App Home
- Open Dev Tools > Application Tab > Cookies -> __session cookie is NOT set

## Expected Behavior
__session cookie must be set with the correct session token value on successful authentication using Google OAuth

### LINKED ISSUE #44: [Bug][Auth] Login component does not consume redirect URL
As a user, I expect to be redirected back to the page I originally intended to visit before I was automatically redirected to the login page.

## Current Behavior
User is always redirected back to home page post login

## Steps to reproduce
- Click on Account in the Nav to visit the Account Page as a logged out user
- Notice automatic redirect to Login Page
- Login into your Meet Near Me account
- Notice you get redirected to home page instead of Accounts page after successful authentication

## Expected Behavior
User to be redirected back to the page they originally intended to visit before automatic redirect to the login page

### LINKED ISSUE #45: [User Story] Emit AWS App hostname to .env
As a developer, I need access to the dynamic AWS URL generated from the SST setup to use in the app for uses like building the redirect URL.

A.C.
 - Emit AWS hostname to .env
 - Replace all hardcoded uses of the hostname with the variable in .env

brianfeister:
This PR adds 

- [x] Google maps embed integration / route for indirect lat /long querying
- [x] Introduces a new `SeshuSessions` dynamodb table that includes a TTL attribute, such that sessions (these database entries are huge and regularly approach the `400kb` db item size limit imposed by `Dynamo) clean themselves up automatically after 24 hrs
- [x] Introduces the concept of `EventValidations` on a `SeshuSession`, this design aims for securing our backend from malicious injection by only allowing the user to set `EventValidations`, which is an array of boolean arrays (e.g. `[ [true, false, true, false], [true, true, true, true] ]`). This means the events themselves (`SeshuSession.EventCandidates`) cannot be modified by the end user, thus only the `EventValidations` array can be mutated, which ensures the user is only capable of messing up their own Seshu ingestion job by manipulating the input
- [ ] TODO: In the next PR we will create a "handoff" task that uses something like `goquery` to query the HTML for matches from positively identified `EventCandidates` and form a DOM querystring that can repeatedly ingest (without the help of OpenAI API) the events that are confirmed as correct per the ingestion schema restrictions

Fixes #45 
Fixes #33

Note that going with Zitadel instead of Clerk also fixes these, though we'll have to fix them if we ever decide to go with Clerk after all

Fixes #42 
Fixes #43
Fixes #44


functions/gateway/handlers/data-handers.go
```diff
@@ -19,40 +19,114 @@ func CreateEvent(w http.ResponseWriter, r *http.Request, db *dynamodb.Client) ht
...
+	}
+
+	// TODO: consider log levels / log volume
+	log.Printf("Inserted new item: %+v", res)
+
+	// TODO: Replace JSON response with htmx template with event data
+	return transport.SendHtmlRes(w, []byte(string(json)), http.StatusCreated, nil)
+}
+
+func CreateSeshuSession(w http.ResponseWriter, r *http.Request, db *dynamodb.Client) http.HandlerFunc {
```
Brandon-G-Tripp, line 56: The seshu session is the wrapper for both the scraping bee and open ai functionality correct? This is kind of the main entry for all of the subsequent work?

functions/gateway/handlers/data-handers.go
```diff
@@ -19,40 +19,114 @@ func CreateEvent(w http.ResponseWriter, r *http.Request, db *dynamodb.Client) ht
...
+}
+
+func UpdateSeshuSession(w http.ResponseWriter, r *http.Request, db *dynamodb.Client) http.HandlerFunc {
+	ctx := r.Context()
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		return transport.SendHtmlRes(w, []byte(string("Error reading request body")), http.StatusUnprocessableEntity, err)
+	}
+	var updateSehsuSession services.SeshuSessionUpdate
+	err = json.Unmarshal([]byte(body), &updateSehsuSession)
```
Brandon-G-Tripp, line 101: @brianfeister It looks like misspelling on this variable updateSeshuSession

functions/gateway/handlers/data-handers.go
```diff
@@ -19,40 +19,114 @@ func CreateEvent(w http.ResponseWriter, r *http.Request, db *dynamodb.Client) ht
...
+	// TODO: consider log levels / log volume
+	log.Printf("Inserted new item: %+v", res)
+
+	// TODO: Replace JSON response with htmx template with event data
+	return transport.SendHtmlRes(w, []byte(string(json)), http.StatusCreated, nil)
+}
+
+func CreateSeshuSession(w http.ResponseWriter, r *http.Request, db *dynamodb.Client) http.HandlerFunc {
+	ctx := r.Context()
+	var createSehsuSession services.SeshuSessionInput
```
Brandon-G-Tripp, line 58: Same here for the var.

functions/gateway/handlers/partial-handlers.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if err != nil {
+		return transport.SendHtmlRes(w, []byte("Invalid JSON payload"), http.StatusUnprocessableEntity, err)
+	}
+
+	latFloat, err := strconv.ParseFloat(lat, 64)
+	if err != nil {
+		return transport.SendHtmlRes(w, []byte("Invalid latitude value"), http.StatusUnprocessableEntity, err)
+	}
+	updateSehsuSession.LocationLatitude = latFloat
```
Brandon-G-Tripp, line 123: Misspelling here as well for Seshu.

functions/gateway/handlers/partial-handlers.go
```diff
@@ -0,0 +1,249 @@
...
+	lonFloat, err := strconv.ParseFloat(lon, 64)
+	if err != nil {
+		return transport.SendHtmlRes(w, []byte("Invalid longitude value"), http.StatusUnprocessableEntity, err)
+	}
+	updateSehsuSession.LocationLongitude = lonFloat
+	updateSehsuSession.LocationAddress = address
```
Brandon-G-Tripp, line 129: Here as well.

functions/gateway/services/seshu_service.go
Brandon-G-Tripp, line 1: @brianfeister This looks like most of the work from the part 1 seshu PR. Are there any notable differences in part 2 to take a look at for the seshu_service.

functions/gateway/handlers/data-handers.go
brianfeister: That's correct. It gets called from within `functions/lambda/go/seshu/main.go` just trying to keep more of the logic in Gateway if possible, and only that one big long-running function in `seshu/main.go`

functions/gateway/handlers/data-handers.go
brianfeister: Good catch. Fixed

brianfeister: Fixed on #50 that is

functions/gateway/services/seshu_service.go
brianfeister: Part 2 adds the whole concept of the `SeshuSessions` table and database persistence which is absent from `feature/seshu`

brianfeister: Actually the changes are listed in the PR description

https://github.com/meetnearme/api/pull/51#issue-2397995541

functions/gateway/handlers/data-handers.go
brianfeister: https://github.com/meetnearme/api/pull/50/commits/9338d1daf73eb2b8f3c5c3210e1ca75e53a7e0ef

functions/gateway/handlers/data-handers.go
brianfeister: https://github.com/meetnearme/api/pull/50/commits/9338d1daf73eb2b8f3c5c3210e1ca75e53a7e0ef

functions/gateway/handlers/partial-handlers.go
brianfeister: https://github.com/meetnearme/api/pull/50/commits/9338d1daf73eb2b8f3c5c3210e1ca75e53a7e0ef

functions/gateway/handlers/partial-handlers.go
brianfeister: https://github.com/meetnearme/api/pull/50/commits/9338d1daf73eb2b8f3c5c3210e1ca75e53a7e0ef

PR #54, opened 2024-07-12
Add Go Lambda Test Coverage and Documentation

Brandon-G-Tripp:
# Add Test Coverage and Documentation for Go Lambdas

Add comprehensive test coverage and documentation for Go Lambda functions

## Description

This PR introduces extensive test coverage for our Go Lambda functions, including unit tests, integration tests, and mocks for DynamoDB. It also adds documentation for our testing approach and updates our CI/CD pipeline to enforce a minimum 50% test coverage threshold.

## Changes

- [x] Implemented table-driven tests for event services package
- [x] Created and refactored to use DynamoDBAPI interface for better testability
- [x] Added tests and improved coverage for the helpers package
- [x] Refactored data handler wrapper for testing and added tests for the data handler
- [x] Implemented table-driven tests for the router along with refactoring of the DynamoDBAPI client
- [x] Added checks for 50% test coverage in CI/CD pipeline
- [x] Created comprehensive testing documentation with examples for our Go Lambda functions

## Testing

- [x] Implemented unit tests for individual functions in helpers, indexing, and services packages
- [x] Added integration tests for handlers and services interaction
- [x] Created mocks for DynamoDB client to facilitate testing without actual database connections
- [x] Ensured all new tests pass locally and in CI environment
- [x] Verified that code coverage meets or exceeds the 50% threshold

## Documentation

- [x] Created a detailed TESTING.md file with guidelines on running tests, writing new tests, and checking code coverage

## Checklist

- [x] Code follows the project's coding conventions and style guidelines
- [x] Code is properly formatted and linted
- [x] Tests have been added to cover the changes and existing functionality
- [x] Documentation has been updated to reflect the changes and new testing approach
- [x] Commits are squashed into logical units with descriptive commit messages
- [x] No unintended changes are included in the pull request

## Additional Information

This PR significantly improves our testing infrastructure and sets a solid foundation for maintaining high code quality as we continue to develop our Lambda functions. The new tests cover various scenarios, including happy paths, error cases, and edge cases.

The addition of the DynamoDBAPI interface allows for easier mocking in tests and improves the overall testability of our codebase. The table-driven tests provide a clear and extensible way to test multiple scenarios for each function.

The new CI/CD check for 50% code coverage ensures that we maintain a good level of test coverage as the project evolves. This will help catch potential issues early and provide confidence in our deployments.

- **Create JSON schema for the event model as source of truth in the SST stacks and in Go API**
- **Move of the event_schema.json file to internal/models dir**
- **Go Event struct to model data for JS and dynamodb**
- **Update Event struct field for ZOrderIndex to []byte**
- **Implementation of the ZOrderIndexing and the mapFloatToSortableInt functions in the indexing package**
- **Tweak / refine daisy ui theming, brand colors, add component for nested checkbox dropdown**
- **Add z order index into the event select and event insert structs along with lat and lon fields.**
- **Add fields to storage stack for lat, lon and z index.**
- **Update the GetEvents function to handle start and end time in the QueryInput**
- **feat(z_order_index): Add latitude and longitude to the event service**
- **Add dev-remote-db script command in package.json**
- **Update date format helpers to use time.RFC3339**
- **feat(z_order_indexing): Add indexing testing and logic for z order index**
- **feat(z_order_index): InsertEvent in events_services to calculatre z order index**
- **feat(z_order_indexing): change lat/long to 32 bits on EventSelect and EventInsert**
- **feat(z_order_indexing): func to get events by z order range (lat, long, start time and appended creation time)**
- **feat(z_order_indexing): z order index calculation and index creation**
- **feat(z_order_indexing): adding the z order indexing to HomePage handler**
- **Removal of logging in events_service.go**
- **feat(local_dynamodb_flag): rename of db helper to IsRemoteDB**
- **feat(local_dynamodb_flag): add .env flag checking for the use of local dynamodb instance versus the AWS resource version**
- **Renaming test file from `z_order_indexing_testing.go` => `z_order_indexing_test.go` so that `go test` properly executes**
- **Attempt to add go testing PR check**
- **run these tests on all branches**
- **run on all pull requests and pushes to all branches**
- **needs to be added from main**
- **Add `go.yml` github actions workflow, and remove unused go import that was failing Go binary build**
- **Revert "Add `go.yml` github actions workflow, and remove unused go import that was failing Go binary build"**
- **Add Go build and test step to Github actions**
- **Update event POST / GET API path**
- **upstream merge from `main`**
- **fix yml syntax**
- **fix github actions checkout version**
- **use recursive search for `go get ./...`**
- **fix unused import failing go binary builds**
- **fix(z_order_indexing): removing unused imports in the z_order_indexing.go**
- **fix(z_order_indexing): correct calculation of the float to sortable integer conversion, use 32 bit integers rather than 64 bit to save space.**
- **fix(z_order_indexing): simply the logic to convert directly to 32 binary string in the z order indexing function.**
- **Documentation(z_order_index): add documentation to the README for use the .env file for the USE_REMOTE_DB boolean flag. This allows a true value for using AWS versus false for using the docker dynamodb**
- **fix(z_order_index): cleanup logs in the events_service.go**
- **fix(remote_db_flag): add remote db flag to the .env.example file**
- **feat(go_lambda_test_coverage): add table driven tests for event services package. This includes dynamodb api mocks for the testing environment.**
- **feat(go_lambda_test_coverage): Creation and refactor to use DynamoDBAPI interface**
- **feat(go_lambda_test_coverage): add tests and coverage of helpers package**
- **feat(go_lambda_test_coverage): refactor for data handler wrapper for testing and tests for the data handler**
- **feat(go_lambda_test_coverage): table driven tests for router along with refactor of the dynamodbapi client**
- **feat(go_lambda_test_coverage): Add checks for 50% test coverage and prevent check from passing if not covered**
- **feat(go_lambda_test_coverage): testing documentation and examples for our go lambda**


brianfeister:
@Brandon-G-Tripp I have no idea why the branches (like this one) stopped running tests, so let's consider that a TODO: that we want to resolve here in this PR.

Essentially, we need to merge `main` back into your branch, which will pull in tons of changes from me (and blow up 🤯 ) your 50% coverage. But if we're testing UI template code and the new routes, much can be covered by just rendering those new templates in a test somewhere

So, I'd love to see a passing "Go test and build" action happen on this branch, then we can merge it

.github/workflows/go.yml
```diff
@@ -32,3 +33,24 @@ jobs:
...
+            echo "Coverage is below 50%"  
+            exit 1
+          fi
+
+      - name: Upload coverage report
+        uses: actions/upload-artifact@v4
+        with: 
+          name: coverage-report
+          path: coverage.txt
+
```
brianfeister, line 56: There's this chicken and egg problem with Github Actions and I can't believe there's no way around it -- changes like this must **already be present** in the target branch for the PR to run the stuff you've got here.

So we might need a first PR that just merges this, then a second PR that benefits from the Github Actions change and runs CI

Consider this permission to go cowboy 🤠 and commit straight to `main` (**but only commit `.github/*` stuff straight to `main`** ) so that you can figure out getting test to run

brianfeister: Also, reminder to pull `main` into your branch here. Bunch of work needed to reconcile merge conflicts

functions/lambda/handlers/page-handlers.go
```diff
@@ -8,14 +8,14 @@ import (
...
 	"github.com/meetnearme/api/functions/lambda/helpers"
 	"github.com/meetnearme/api/functions/lambda/services"
 	"github.com/meetnearme/api/functions/lambda/templates/pages"
 	"github.com/meetnearme/api/functions/lambda/transport"
+	"github.com/meetnearme/api/functions/lambda/types"
 )
 
-func GetHomePage(ctx context.Context, r transport.Request, db *dynamodb.Client) (transport.Response, error) {
+func GetHomePage(ctx context.Context, r transport.Request, db types.DynamoDBAPI) (transport.Response, error) {
     // Extract parameter values from the request query parameters
```
brianfeister, line 19: When I run your branch locally, I get this error. Do you know if there's something I need to do?

```
|  Invoked functions/lambda
|  +1ms    2024/07/12 06:35:13 {"errorMessage":"handlers may not take more than two arguments, but handler takes 3","errorType":"errorString"}
|  Error: handlers may not take more than two arguments, but handler takes 3
```

brianfeister: @Brandon-G-Tripp beautiful ✨ ... got it running, had to delete my local dev AWS resources as usual

image

image

.github/workflows/go.yml
```diff
@@ -32,3 +33,24 @@ jobs:
 
       - name: Run Go tests
         run: go test ./...
+
+      - name: Run tests and generate coverage
+        run: |
+          go test -v -coverprofile=coverage.out ./functions/...
+          go tool cover -func=coverage.out > coverage.txt
+
```
brianfeister, line 41: Are we expecting to run these locally? If so, should we add them to `.gitignore`? 

image

functions/lambda/helpers/helpers_test.go
```diff
@@ -0,0 +1,105 @@
+package helpers
+
+import (
+    "os"
+    "testing"
+)
+
+func init() {
+    os.Setenv("GO_ENV", "test")
```
brianfeister, line 9: Can this be set in `TestMain` to avoid boilerplate and having to remember to do this in the `init` block of very new test file we create?

functions/lambda/helpers/helpers_test.go
```diff
@@ -0,0 +1,105 @@
...
+    for _, tt := range tests {
+        t.Run(tt.name, func(t *testing.T) {
+            os.Setenv("USE_REMOTE_DB", tt.useRemoteDB)
+            os.Setenv("SST_STAGE", tt.sstStage)
+            result := IsRemoteDB()
+            if result != tt.expectedResult {
+                t.Errorf("IsRemoteDB() = %v, want %v", result, tt.expectedResult)
+            }
+        })
+    }
```
brianfeister, line 78: Nice 👍  Good to cover this, even though it seems irrelevant, as it could impact production if someone messes with the logic in this helper function

functions/lambda/handlers/data_handlers_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/meetnearme/api/functions/lambda/services"
+	"github.com/meetnearme/api/functions/lambda/transport"
+	"github.com/meetnearme/api/functions/lambda/types"
+)
+
+// Mock DynamoDB client
+type mockDynamoDBClient struct {
```
brianfeister, line 18: I think once you merge `main` into your branch, you'll find that my new pattern for accessing the `Db` using a singleton approach rather than mucking with increasingly over-dependent route handlers, means that you'll want to globalize this type to be re-usable in the various services that pull the `Db` instance in. I could be wrong though, just something to think about

functions/lambda/handlers/data_handlers_test.go
```diff
@@ -0,0 +1,146 @@
...
+            },
+            expectedStatus: http.StatusInternalServerError,
+            expectedBodyCheck: func(body string) error {
+                if body != "Internal Server Error" {
+                    return fmt.Errorf("expected 'Internal Server Error', got '%s'", body)
+                }
+                return nil
+            },
+        },
+    }
```
brianfeister, line 116: Beautiful! ❤️

functions/lambda/main.go
```diff
@@ -4,26 +4,31 @@ import (
...
 	"github.com/aws/aws-lambda-go/lambda"
-	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
 	"github.com/meetnearme/api/functions/lambda/handlers"
 	"github.com/meetnearme/api/functions/lambda/transport"
+	"github.com/meetnearme/api/functions/lambda/types"
 )
 
 var router *transport.Router
-var db *dynamodb.Client
+var db types.DynamoDBAPI
```
brianfeister, line 13: 👍

functions/lambda/main.go
```diff
@@ -4,26 +4,31 @@ import (
...
 
-func init() {
-	db = transport.CreateDbClient()
+func InitializeApp(dbClient types.DynamoDBAPI) {
```
brianfeister, line 15: Nice work 📈

functions/lambda/main.go
```diff
@@ -4,26 +4,31 @@ import (
...
 }
 
-func Router(ctx context.Context, req transport.Request) (transport.Response, error) {
-	return router.ServeHTTP(ctx, req, db)
+func init() {
+    InitializeApp(transport.CreateDbClient())
+}
+
+func Router(ctx context.Context, req transport.Request, dbClient types.DynamoDBAPI) (transport.Response, error) {
+	return router.ServeHTTP(ctx, req, dbClient)
```
brianfeister, line 31: Whoops! Made a bunch of new work for you in my branch, sorry about that! 🙇 Let me know if you need help thinking through merging this with the gorilla/mux + auth router changes

functions/lambda/main_test.go
```diff
@@ -0,0 +1,138 @@
...
+			},
+			expectedStatus: 200,
+            checkBody: func(body string) bool {
+                return strings.Contains(body, "Meet Near Me - Login") && strings.Contains(body, "Login Page")
+            },
+		},
+		{
+			name: "GET /events/123",
+			request: transport.Request{
+				RequestContext: events.APIGatewayV2HTTPRequestContext{
```
brianfeister, line 72: This one is another big refactor! I've migrated to use `transport.SendHtmlRes` and `transport.SendServerRes` everywhere instead of this AWS event directly

functions/lambda/services/event_service_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+    if insertedEvent.Name != createEvent.Name {
+        t.Errorf("Expected inserted event Name to be %s, got %s", createEvent.Name, insertedEvent.Name)
+    }
+
+    if len(insertedEvent.ZOrderIndex) == 0 {
+        t.Error("Expected ZOrderIndex to be set")
+    }
+
+    expectedZOrder, err := indexing.CalculateZOrderIndex(eventTime, createEvent.Latitude, createEvent.Longitude, "normal")
```
brianfeister, line 135: This especially is great to see being tested. It's fine to increase test coverage, but what we care about is careful and well-thought out decisions about WHERE to test, since we won't have time to test everywhere

brianfeister (approved):
I actually think this looks great and is ready for merge once you resolve the merge conflicts.

Well done! 🏆

functions/lambda/helpers/helpers_test.go
Brandon-G-Tripp: Yeah that is a great idea. I will open an issue and do a follow up PR. The challenge is that the TestMain is bound per package. Documenting here but I saw making the test helper package and run through a sync. Not real clear on this pattern but we could do that to initialize for all the tests running in parallel.

.github/workflows/go.yml
Brandon-G-Tripp: My thought is that these are run in CI and then the coverage.txt artifact can be uploaded if we want at a later point.

.github/workflows/go.yml
Brandon-G-Tripp: Yeah I can commit the workflow additions. I know I have worked with the workflow files a bunch lately on some projects and it is terrible there is not a good way to test the action below deploying.

Brandon-G-Tripp:
@brianfeister Once I get all of these conflicts resolved and work the testing back into our code. It would be helpful to pair on testing the UI by rendering as suggested. I will give it an attempt first to have a baseline and then we can discuss the best approach for our UI.

brianfeister (changes requested):
Whoops, after re-deploying / nuking all my AWS assets, still seeing this, which is a blocker for merging. lmk if you need help reproducing -- al I did was visit the apex URL (API) of the local deployment

```
[sst] |  Error: handlers may not take more than two arguments, but handler takes 3
[sst] |  Invoked functions/lambda
[sst] |  +2ms    2024/07/12 20:31:24 {"errorMessage":"handlers may not take more than two arguments, but handler takes 3","errorType":"errorString"}
[sst] |  Error: handlers may not take more than two arguments, but handler takes 3
```

Brandon-G-Tripp:
It looks my checks are failing from the templ.go file. I corrected the file but not sure how to get it to regenerate the .go file from the updated templ. Is that in CI or do I have to trigger it some how?

.github/workflows/go.yml
brianfeister: I think that future-looking plan doesn't change that we want to add these files to `.gitignore` if we upload them, that will be a specific github actions scripted thing, it won't hurt for them to be `.gitignored` and we **definitely** want them ignored for the scenario where we run this locally

.github/workflows/go.yml
```diff
@@ -32,3 +33,24 @@ jobs:
...
+      - name: Check coverage
+        run: |
+          COVERAGE=$(grep -Po '(?<=total:\s{6})\d+\.\d+' coverage.txt)
```
brianfeister, line 44: I think this same command was causing issues in your `go.yml` PR here

https://github.com/meetnearme/api/pull/81#pullrequestreview-2202898709

functions/gateway/helpers/db_helpers.go
```diff
@@ -18,11 +19,19 @@ func IsRemoteDB() bool {
 
 func GetDbTableName(tableName string) string {
 	// this must be added in stacks/ApiStack.ts and stacks/StorageStack.ts
-	var SST_Table_tableName = os.Getenv("SST_Table_tableName_" + tableName)
+	var SST_Table_tableName_Events = os.Getenv("SST_Table_tableName_" + tableName)
```
brianfeister, line 22: I believe this is a regression. We now have multiple SST DB tabes, therefore, we have constants for the suffixes like `Events` and `SeshuSession` this regresses the improvement I made to parameterize that

brianfeister: Have you verified that `/admin/add-event-source` still works from a purely manual / functional test perspective? I expect this change would break it

functions/gateway/helpers/db_helpers.go
```diff
@@ -18,11 +19,19 @@ func IsRemoteDB() bool {
...
 	}
-
-	return SST_Table_tableName
+    if os.Getenv("GO_ENV") != "test" {
+        log.Printf("Log Get Db Table: %v", SST_Table_tableName_Events)
+    }
+    if SST_Table_tableName_Events == "" {
+        return ""
+    }
+	return SST_Table_tableName_Events
```
brianfeister, line 35: Breaks Seshu, I assume, as mentioned above

functions/gateway/handlers/data_handlers.go
```diff
@@ -9,45 +10,70 @@ import (
...
-	ctx := r.Context()
+    return func(w http.ResponseWriter, r *http.Request) {
+        db := Db
+        if db == nil {
+            db = transport.GetDB()
+        }
+        createEventHandler(r.Context(), w, r, &RealEventService{}, db)
+    }
+}
+
```
brianfeister, line 40: What about the singleton pattern we discussed where we were going to use `TestMain` for this? I believe that's **way** less boilerplate than having a wrapper function inside every handler.

This definitely feels brittle to me and if the `TestMain` approach is not valid for some reason, I believe we need some "third way" (e.g. not this way, not the `TestMain` way if those are the only two options on the table so far)

brianfeister: An ideal solution here would accept the `test` flag in the runtime start command and **ONLY** introduce branching logic in this one DB init function

https://github.com/meetnearme/api/pull/54/files#diff-39e602ef683ead3def246a979ee1880ade6c841952f3a161da36d92cd67d7e42R71-R88

brianfeister: Another possible solution: Why is there not an AWS Dynamo Go "testing" SDK (I believe it's part of the default package maybe?) we should be able to use that and mock so that calling an instance of the AWS Dynamo SDK instead instantiates a Dynamo Mock. I think it's possible that everything in `test_hepers` https://github.com/meetnearme/api/pull/54/files#diff-256af9b7cfaea1643572450fca6a91ff62f3e4baeb09411a254215fa058f5fb9R1 is actually doing something that the AWS SDK already provides and we're not using. At least I **hope** that's the case. Thoughts? 

Looking at it closer, maybe there's no mocking SDK function. 

Note how in this example, the test code for the `ItemGetter` here https://github.com/aws/aws-sdk-go/blob/main/example/service/dynamodb/unitTest/unitTest.go#L24 does not leak into the implementation, but lives only in the test file with it's special getter definition here: https://github.com/aws/aws-sdk-go/blob/main/example/service/dynamodb/unitTest/unitTest_test.go#L36

So, the advantage here is if I change the behavior of the method, it's the **TEST** (with it's mock function interface) that would break. The way you have it now, it's reversed and it's the **implementation** that's at risk for breaking if things are changed, we want the **tests** to break while the implementation stays healthy if I make changes to it

brianfeister: I think this is the critical piece. Instantiate the `ItemGetter` (our handler) and then overload the internal reference `getter.DynamoDB` on the instance (think JavaScript / Java classes here)

https://github.com/aws/aws-sdk-go/blob/main/example/service/dynamodb/unitTest/unitTest_test.go#L39-L40

We can keep our singleton pattern and still overload just the `<method>.DynamoDB` property on each class (though we would have to store `<method>.DynamoDB` which we don't yet

brianfeister (changes requested):
I need to review more later, but I'm concerned about the amount of boilerplate. Let's think through / discuss this a bit more

functions/gateway/handlers/data_handlers.go
Brandon-G-Tripp: I ran into this issue earlier on Z index stuff because it seems that the SDK used to have a mock for the Dynamodb api but I believe it was removed in the latest version of the SDK that we are currently on. 

Yeah I definitely get that worry of too much boilerplate. I think it is a good time to hash out our testing pattens so we are not having to play catchup in the future. The whole process felt pretty clunky and there has to be something cleaner. I wonder if this is a point where it might make sense to introduce a library for mocking and check if there is a library that works for AWS types and resources.

Brandon-G-Tripp: Many of my attempts involved trying to overload the internal reference but I kept running into compile issues that would not allow me to reassign those values from anywhere outside the package. Maybe we can internalize it to the package though with the singleton pattern.

functions/gateway/helpers/db_helpers.go
Brandon-G-Tripp: Ok I can revert this. Where is the work you did again for the parameterization?

functions/gateway/handlers/main_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/meetnearme/api/functions/gateway/transport"
+)
+
+func TestMain(m *testing.M) {
+	log.Println("Setting up test environment for handlers package")
+
+	// Set GO_ENV to "test" to trigger test-specific behavior
+	os.Setenv("GO_ENV", "test")
+
+	mockDB := &test_helpers.MockDynamoDBClient{
```
msnorth, line 21: thought going forward: we will need to update this if we are switching to a ddbmapper

msnorth (approved):
Ship it

brianfeister (approved):
LGTM

functions/gateway/helpers/utils.go
```diff
@@ -36,30 +33,4 @@ func GetBaseUrlFromReq(r *http.Request) string {
...
-							PrintContextInternals(reflectValue.Interface(), true)
-					} else {
-							fmt.Printf("field name: %+v\n", reflectField.Name)
-							fmt.Printf("value: %+v\n", reflectValue.Interface())
-					}
-			}
-	} else {
-			fmt.Printf("context is empty (int)\n")
-	}
-}
```
brianfeister, line 65: Deleted this, moved as our first "snippet" to this doc:
https://docs.google.com/document/d/1EC5jUTC0qe5kIPo5vlAtuJwjdMDjbvkyUl1d3YVd-O4

functions/gateway/helpers/utils_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+					result := FormatTime(tt.input)
+					if result != tt.expected {
+							t.Errorf("FormatTime(%q) = %q, want %q", tt.input, result, tt.expected)
+					}
+			})
+	}
+}
```
brianfeister, line 53: These are simply moved from `helpers/helpers_test.go` to `helpers/utils_test.go` to match the convention of the test file being named identically to the go application code filename

functions/gateway/helpers/utils_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+					result, _ := TruncateStringByBytes(tt.input1, tt.input2)
+					if result != tt.expected {
+							t.Errorf("TruncateStringByBytes(%q) = %q, want %q", tt.input1, result, tt.expected)
+					}
+			})
+	}
+}
```
brianfeister, line 73: Added for increased coverage. Coverage threshold was failing after the many extraneous `log` statements were removed

brianfeister (approved):

PR #56, opened 2024-07-13
Cleanup Seshu form UI 🍴 

brianfeister:
This PR is purely UX-based, improving error messaging and handling as well as state management for the `/admin/add-event-source` route.


functions/gateway/handlers/page_handlers.go
```diff
@@ -184,7 +184,7 @@ func GetEventDetailsPage(w http.ResponseWriter, r *http.Request) http.HandlerFun
 	return transport.SendHtmlRes(w, buf.Bytes(), http.StatusOK, nil)
 }
 
-func GetAdminPage(w http.ResponseWriter, r *http.Request) http.HandlerFunc {
+func GetAddEventSourcePage(w http.ResponseWriter, r *http.Request) http.HandlerFunc {
```
brianfeister, line 187: `Admin` was a placeholder, updating the function naming

functions/gateway/handlers/partial_handlers.go
```diff
@@ -64,7 +64,7 @@ func GeoLookup(w http.ResponseWriter, r *http.Request) http.HandlerFunc {
 		return transport.SendHtmlRes(w, []byte(string("Error getting geocoordinates: ") + err.Error()), http.StatusInternalServerError, err)
 	}
 
-	geoLookupPartial := partials.GeoLookup(lat, lon, address, false)
```
brianfeister, line 67: The boolean logic for switching in the template isn't needed, simplifying since we always respond with `Location Confirmed`

functions/gateway/handlers/partial_handlers.go
```diff
@@ -195,7 +195,7 @@ func SubmitSeshuEvents(w http.ResponseWriter, r *http.Request) http.HandlerFunc
 func SubmitSeshuSession(w http.ResponseWriter, r *http.Request) http.HandlerFunc {
 
 	ctx := r.Context()
-	var inputPayload SeshuSessionEventsPayload
+	var inputPayload SeshuSessionSubmitPayload
```
brianfeister, line 198: This was a copy / paste error and should have never been `SeshuSessionEventsPayload`

functions/gateway/templates/pages/add_event_source.templ
```diff
@@ -30,7 +30,7 @@ templ AdminPage() {
 		</div>
 		<div id="event-source-container" class="col-span-5 card border-2 border-base-300 bg-base-200 p-10 rounded-box">
 			<h2 class="text-3xl font-bold ">Add a Target URL</h2>
-			<form class="group" novalidate hx-post={ templ.EscapeString(os.Getenv("SESHU_FN_URL")) } hx-ext="json-enc" hx-target="#event-candidates-inner" hx-disabled-elt="input[name='url'], button[type='submit']" @submit.prevent="formStep < 2 ? incrementFormStep() : ''" @htmx:after-request="updateHasEvents(event)">
+			<form class="group" novalidate hx-post={ templ.EscapeString(os.Getenv("SESHU_FN_URL")) } hx-ext="json-enc" hx-target="#event-candidates-inner" hx-disabled-elt="input[name='url'], button[type='submit']" @submit.prevent="formStep < 2 ? handleURLSubmit() : ''" @htmx:after-request="handleUrlRes(event)">
```
brianfeister, line 33: Doing more than one thing in the `handle*` functions, so moved them to the `<script>` below

functions/gateway/templates/pages/add_event_source.templ
```diff
@@ -92,12 +92,21 @@ templ AdminPage() {
 				<div id="event-candidates-location" class="has-toggleable-text">
 					<h3 class="text-2xl font-bold mt-16">Are all events at the target URL located in the same geography?</h3>
 					<label class="label cursor-pointer justify-normal">
-						<input id="event-candidates-location-toggle" type="checkbox" class="toggle toggle-success mr-4" checked/>
+						<input x-model="hasFallbackLocation" id="event-candidates-location-toggle" type="checkbox" class="toggle toggle-success mr-4" />
```
brianfeister, line 95: Need to track this value at the Alpine form state level

functions/gateway/templates/pages/add_event_source.templ
```diff
@@ -92,12 +92,21 @@ templ AdminPage() {
...
-							<form class="group" novalidate hx-patch="/api/html/seshu/session/location" hx-ext="json-enc" hx-target="#location-confirmation" hx-disabled-elt="input[name='location'], button[type='submit']" @submit.prevent="formData.address ? '' : ''">
+							<form
+								class="group"
+								novalidate
+								hx-patch="/api/html/seshu/session/location"
+								hx-ext="json-enc"
+								hx-target="#location-confirmation"
+								hx-disabled-elt="input[name='location'], button[type='submit']"
+								@submit.prevent="formData.address ? '' : ''"
+								@htmx:after-request="handleLocationValidation(event)"
```
brianfeister, line 108: This is the main change, since we need to change formState (`this.locationValidated`) only if successful

functions/gateway/templates/pages/add_event_source.templ
```diff
@@ -122,8 +131,9 @@ templ AdminPage() {
 								></iframe>
 								<br/>
 								<br/>
-								<p>Is the location in the map above accurate? (enter something in the text field above before adding)</p>
-								<button type="submit" :disabled="!formData.address?.length || !formData.url?.length" class="btn btn-primary mt-3"><svg xmlns="http://www.w3.org/2000/svg" class="stroke-current h-8 w-8" fill="none" viewBox="0 0 24 24"><path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M9 12l2 2 4-4m6 2a9 9 0 11-18 0 9 9 0 0118 0z"></path></svg> Add Location <span class="htmx-indicator loading loading-spinner loading-sm"></span></button>
+								<button type="submit" :disabled="!formData.address?.length || !formData.url?.length" class="btn btn-primary mt-3">
```
brianfeister, line 134: It is now (finally!) not possible to submit the form if the various permutations of valid state are not satisfied. They are:

1. MUST have a URL
2. MUST confirm events are really events (after toggling the green / red event cards)
3. MUST enter a string of some kind to get a map via Google map query
4. MUST confirm Location, BUT ONLY IF the toggle is checked (starts off checked by default because we want to encourage fallback location to be able to error on the side of pulling in more events)

If all of the above are satisfied, the submit button should be enabled, if not, should be disabled

functions/gateway/templates/pages/add_event_source.templ
```diff
@@ -132,20 +142,36 @@ templ AdminPage() {
...
-					<div id="final-submission" class="mt-5 htmx-hide-in-flight" aria-live="polite">
-						<div class="skeleton card card-compact h-24 w-full htmx-show-in-flight"></div>
+				<form
+					class="group"
+					novalidate
+					hx-post="/api/html/seshu/session/submit"
+					hx-ext="json-enc"
+					hx-target="#final-submission"
+					hx-disabled-elt="button[type='submit']"
+					@htmx:after-request="handleFinalSubmission(event)"
```
brianfeister, line 156: Sets the `formStep` to 3, highlighting the top left final step

functions/gateway/templates/pages/add_event_source.templ
```diff
@@ -132,20 +142,36 @@ templ AdminPage() {
...
+				>
+					<input
+						x-model="formData.url"
+						name="url"
+						id="url"
+						type="hidden"
+					/>
```
brianfeister, line 163: This is an HTMX quirk, since the Go API Gateway calls for these HTMX endpoints require `url` as part of the payload, the default behavior pre-React days was to include that as part of a named input field in the form (`<input name="url" />` in this case.

We leverage Alpine's `x-model` to always keep the value in sync, but we make the input field hidden in all cases aside from the first input field where users enter the value

functions/gateway/templates/pages/add_event_source.templ
```diff
@@ -132,20 +142,36 @@ templ AdminPage() {
...
+					<button
+						:disabled="!hasEvents || !eventsValidated || (hasFallbackLocation && !locationValidated) || formStep < 2"
```
brianfeister, line 165: The conditions that satisfy the enabling of the submit button

Brandon-G-Tripp (commented):
@brianfeister Hey since this PR is related to the Seshu UI and following the Zitadel reference I would love to see the flow for adding the callback URL and related values to the .env and also the documentation in the README for setting that up in the Zitadel console.

functions/gateway/templates/pages/add_event_source.templ
Brandon-G-Tripp: Ok I opened an issue for number 2. Small bug in my opinion that may want to be incorporated into this PR. The confirm event data button allows submitting even if all events are shown to be invalid by the user. May need different handling or just a UI change for the message. 

See - #58

Brandon-G-Tripp: Screen Shot 2024-07-13 at 5 32 17 PM

The UI allowed me to click Submit event source even without a string of any kind in the event address field.

Brandon-G-Tripp (changes requested):
Hey everything looks good overall. I just want to confirm the UX flow that did not seem to function correctly in light of the Submit Event Source button. The main change being able to submit the event source without the string for the Google Maps rendering. 

The second item was the ability to confirm the event data even if all events were select as not events and then the need to have a differing message if no events are selected and prevent submission. @brianfeister

brianfeister:
@Brandon-G-Tripp 

> I just want to confirm the UX flow that did not seem to function correctly in light of the Submit Event Source button. The main change being able to submit the event source without the string for the Google Maps rendering.

I see what you mean, the `Confirm Location` button was enabled with no address provided. That's what you're saying right? I've fixed that here https://github.com/meetnearme/api/pull/56/commits/f11da851df313f05bf5c7e28ab0904aa4998fb5f

The bug you opened for #58 is fixed in this commit:
https://github.com/meetnearme/api/pull/56/commits/3b57a2f3ffb8b46fcea51b29f6be6e0c4da236d3

brianfeister:
@Brandon-G-Tripp also, great job on the QA aspect of testing this 🏆 

Noticing those UX issues and then requesting changes was the right thing to do!

functions/gateway/templates/pages/add_event_source.templ
brianfeister: This should be fixed now 👍

Brandon-G-Tripp (approved):
Ok I am satisfied with this. I think everything is good to go for this PR. With the follow up PR approved as well.

PR #59, opened 2024-07-15
Make zitadel auth config dynamic per deployed environment + set APEX_URL

### LINKED ISSUE #55: [bug] [auth] zitadel callback uri is not set in build / bootstrap
Right now the only reason that the zitadel callback URL works is because I deploy it first and then redeploy it once I know the Apex URL. This is a big problem for smooth deployments and needs a dynamic solution that happens at build time

brianfeister:
This PR fixes #55. Also addresses @Brandon-G-Tripp 's concern about adding to the README mentioned in #56 here https://github.com/meetnearme/api/pull/56#pullrequestreview-2176556656

### Changes

- [x] Migrate away from `ZITADEL_REDIRECT_URI` which was non-dynamic and required a re-deploy
- [x] Introduce a way to dynamically get the `APEX_URL` in Go runtime env space
- [x] Add documentation on configuring Zitadel auth 

### Testing this PR

* Visit [the environment for this deployment](https://0sbos6nlc4.execute-api.us-east-1.amazonaws.com/admin/add-event-source) for this feature branch and validate that auth works (with the caveat that you must log in / have an active session here first:  https://meet-near-me-production-8baqim.zitadel.cloud/ui/console/projects/273257176187855242/apps/273257486885118346)
* I've already configured the redirect and post logout URIs for this deployment


Brandon-G-Tripp (approved):
Ok awesome. This looks great and helpful to have eliminated the manual entry of a .env field for development. Thanks!

PR #49, opened 2024-07-02
Proposal for `User` schema (Please discuss 🗣️ !) 

brianfeister:
This was originally proposed by @j4inam on Discord here:
https://discord.com/channels/1096208627192320030/1151217204126290011/1240831524384014366

I have made the following modifications from the original proposal:

1. Dropped `reviews` (I personally would like to avoid this in the long-term. Perhaps it will become necessary but I'd rather not introduce that complexity, I see it as having minimal benefit and making us too much like Yelp, etc)
1. Added `EventSlug[]` because we want events to match URLs without doing an extra DB lookup. That said, we could run into internationalization issues with this, 
1. Replaces `roles` with => `entitlements`. I don't think our use case is enough to need roles (which would be arrays that need merging, slowing page load down) ... I'm of the opinion that we only need "capabilities" since they map 1:1 and allow users to unlock different `entitlements` (capabilities) in different orders without needing an associated role.
1. Added `mutedCurators`. Had a great conversation with someone last night where I realized that the problem of going to the site and always seeing 1,000 Alcoholics Anonymous meetings you're not interested in can be easily solved by being able to "mute" curators. Also helps in the situation where one person views the events posted by another as "inappropriate"
1. Flattened `location` and dropped complex address (Street1 / 2, city, state, zip, country) because I want us to be `latitude` and `longitude` centric. I want to verify on a map (optionally with geolocation in the browser) where the user is located
1. Dropped `password` as this will be in our auth service, not our DB

The typescript type is way easier to read:

```
type ISOString = string;
type UserId = uuid;
type EventSlug = string;

type User {
 id: string;
 firstName: string;
 lastName: string;
 mobile?: string;
 locationLatitiude?: number;
 locationLongitude?: number;
 locationAddress?: string;
 karma: number;
 email: string;
 isEmailVerified: boolean; 
 dateOfBirth: ISOString;
 entitlements: 'ingestEvent' | 'postEvent' | 'editEvent' | 'reportEvent' | 'hasSubdomain' | 'hasBranding'
 isActive: boolean;
 isVerified: boolean;
 badges?: string[]; 
 about?: string;
 avatarURL: string;
 notificationSettings: UserNotificationSettings; 
 createdAt: ISOString;
 lastLoggedIn: ISOString;
 eventsRSVP?: Event[];
 eventsAttended?: Event[];
 eventsSaved?: Event[];
 favoriteCategories?: EventSlug[];
 favoriteCurators?: UserId[];
 mutedCurators?: UserId[];
[...]

PR #62, opened 2024-07-18
[DISCUSSION] Entity Relation Architecture Schema

j4inam:
This PR highlights various entities and relation between them which will be fundamental to app's architecture.

This PR includes information discussed in PRs #48 and #49

**NOTE**: Refer to PR comments and description in #48 and #49


architecture/event.schema.json
```diff
@@ -0,0 +1,125 @@
...
+    },
+    "locationLatitude": {
+      "type": ["number", "null"]
+    },
+    "locationLongitude": {
+      "type": ["number", "null"]
+    },
+    "locationAddress": {
+      "type": ["string", "null"]
+    },
```
brianfeister, line 72: I recently ran into a problem with this, because this field is set to `number` as a type, any invalid / empty value in Dynamo will set this as `0` but for latitude and longitude `0` is a valid value (google "null island") 

I have created a "out of bounds" (in terms of valid lat / lon) placeholder that I insert in `SeshuSession`s so that I can detect "invalid / incorrect" for latitude and longitude.

https://github.com/meetnearme/api/pull/60/files#diff-112a7e21d5f0ffddb8cef87be4c7f05e200404c73101b69a9d1e6bf41afd6b39R104

However, in the case of `event.schema.json` you have here, `null` should not be an option because we actually **never** allow any event that doesn't have a valid `latitude` and `longitude` in that scenario it would be impossible to build the `partitionKey` ("zOrderIndex" we call it, interleaving bytes for lat / lon / startTime / createTime) and so we reject the event and it should never touch the DB

architecture/event.schema.json
```diff
@@ -0,0 +1,125 @@
...
+    "locationCountry": {
+      "type": ["string", "null"]
+    },
+    "locationZipcode": {
+      "type": ["string", "null"]
+    },
```
brianfeister, line 78: I was envisioning that we will always derive location from some pair of geocoordinates. So, if they don't provide any, we'll get it from the ISP headers (via Cloudflare). In terms of normalizing the data, I was going to do only `latitude` / `longitude` / `locationAddress` (which is a full string, that we get back normalized from Google that includes country, postal code, etc. The idea would be that we assume Google is the source of truth for this and we can later derive more data from this string that we know we can use to fetch a stable location from Google with more detail later.

architecture/event.schema.json
```diff
@@ -0,0 +1,125 @@
...
+  },
+  "type": "object",
+  "properties": {
+    "id": {
+      "$ref": "#/definitions/UUID"
+    },
+    "owner": {
+      "$ref": "https://github.com/meetnearme/api/architecture/User.json"
+    },
+    "ingestion": {
```
brianfeister, line 22: This should be a reference, so something like `ingestionID`?

architecture/category.schema.json
```diff
@@ -0,0 +1,34 @@
+{
+  "$id": "https://github.com/meetnearme/api/architecture/Category.json",
```
brianfeister, line 2: I tried to run `json-schema-to-typescript` on the schema files you have here, but the github.com links fail. I suppose they would work after we merge to main? Is there an easy way to do relative `$id`s that only care about the path from the current working directory?

architecture/event.schema.json
```diff
@@ -0,0 +1,125 @@
...
+    }
+  },
+  "required": [
+    "id",
+    "owner",
+    "ingestion",
+    "name",
+    "startTime",
+    "recurrenceRule",
+    "recurrenceId",
```
brianfeister, line 110: `recurrenceRule` is optional, not all events will be recurring. @Brandon-G-Tripp was asking before if the RRULE can be an empty string in the case of none being set. If that's valid, we can leave `recurrenceRule = ''` but `recurrenceId` is definitely optional / nullable, since many events will not be recurring

architecture/event.schema.json
```diff
@@ -0,0 +1,125 @@
+{
```
brianfeister, line 1: Let's add `imageUrl` (unsure if we need to generate multiple thumbnails, etc. I'd rather avoid that complexity at this stage) as we know the work I'm going to do for Brandolon's team creating their own events will include images

brianfeister: Thinking out loud -- in S3, we should plan to have a predictable S3 bucket path `"<root>/events/<id>/featured.jpg"` or something so that we can assume the path and not have to wait for an image to upload, etc. when a person is adding an event.

architecture/event.schema.json
```diff
@@ -0,0 +1,125 @@
...
+  "$id": "https://github.com/meetnearme/api/architecture/Event.json",
```
brianfeister, line 2: Let's also add `price` (defaulting to `0`) and `currency` (defaulting to `USD`)

architecture/user.schema.json
```diff
@@ -0,0 +1,144 @@
...
+          "hasBranding"
+        ]
+      }
+    },
+    "isActive": {
+      "type": "boolean"
+    },
+    "isVerified": {
+      "type": "boolean"
+    },
```
brianfeister, line 75: How do you envision this working exactly? What would constitute verification. I think you're right, and actually, Brandolon's teams events might be our first "verified" since we trust the source, but just paint a picture for me of what a future might look like for handling this beyond our internal team. If someone (we don't know / trust) has "create events" entitlements, how do we know we trust / `isVerified` the event?

architecture/user.schema.json
```diff
@@ -0,0 +1,144 @@
...
+        "enum": [
+          "ingestEvent",
+          "postEvent",
+          "editEvent",
+          "reportEvent",
+          "hasSubdomain",
+          "hasBranding"
+        ]
+      }
+    },
```
brianfeister, line 69: So this one, I failed to realize earlier, probably needs to be handled at the Zitadel auth layer to prevent users from escalating privileges... thoughts?

architecture/eventRSVP.schema.json
```diff
@@ -0,0 +1,39 @@
...
+      "format": "date-time"
+    },
+    "UUID": {
+      "type": "string",
+      "format": "uuid"
+    },
+    "RSVPStatus": {
+      "type": "string",
+      "enum": ["yes", "no", "maybe"]
+    }
```
brianfeister, line 16: This one I'm not sure about. My vision (and I'm open to being wrong here!) was that you click "RSVP" in the UI, and we mark that as "attending" (but more importantly, we open your calendar app and have you add it to your REAL calendar rather than creating yet another annoying new app that you have to check)

Thoughts?

architecture/user.schema.json
brianfeister: Whoops! I see you've got this on the `user`, which probably makes more sense and we can derive `isVerified` for events (probably should add to the `event.schema.json`) from the user, but we should handle that at creation / update, since `event` is in DynamoDB space (so we can't do a table join on `user` inside of `event` queries in a performant way and want to avoid doing so)

architecture/event.schema.json
```diff
@@ -0,0 +1,125 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
brianfeister, line 3: Proabably want to add `isVerified` from `user` (but we require `user` when an event is created / updated, so we just push it to `event` in a non-relational / "push" way)

architecture/event.schema.json
```diff
@@ -0,0 +1,125 @@
...
+  "definitions": {
```
brianfeister, line 4: One thing we might want is `status` for events. I was imagining a canceled event. In the scenario where clicking RSVP as the event to your calendar with a permit in the description back to the event on meet near me, we might want to make it so that an event that gets canceled instead of just changing the date somehow shows to the user that it was canceled so that when they go to the event in their calendar and find it's not happening, they can click the link to get back and see that it's not just deleted but rather canceled by the organizer.

architecture/event.schema.json
```diff
@@ -0,0 +1,125 @@
...
+    "ISOString": {
```
brianfeister, line 5: Also we will need `isExternalEvent` for distinguishing between paid external events (where we don't handle ticketing) and the paid events that are created within the platform

architecture/event.schema.json
```diff
@@ -0,0 +1,131 @@
...
+    "eventCost": {
+      "type": "number"
+    },
+    "recurrenceRule": {
+      "type": "string"
+    },
+    "registrationFields": {
+      "items": {
+        "type": "string"
+      },
```
brianfeister, line 49: I've added this just now, but more work needs to be done here to flesh this out. I plan to use JSON schema validation and offer an array of limited fields that will eventually be user-defined. The `gojsonschema` package looks likely to fit our use case https://github.com/xeipuuv/gojsonschema

An example 

```
{ 
  eventTitle: "Karaoke in DC",
  registrationFields: [
   {
     type: "email",
     displayText: "What is your email?",
     required: true,
   },
   {
     type: "multiSelect",
     displayText: "What evenings work best for you?",
     values: ["mondays", "tuesdays", "wednesdays", "thursdays"]
   }
  ],
  ...
}
  ```

File: architecture/event.schema.json (hunk 13 above)
Comment (brianfeister) [reply architecture/event.schema.json:5]: Actually we need `externalUrl` as well for linking out to the original event URL if it was scraped (we want to maintain this link because there's always a risk that our event data is out of date and if there's confusion, we should help users by maintaining that external pointer so they can see if the listing somehow changed)

File: architecture/event.schema.json [hunk 15]
```diff
@@ -0,0 +1,135 @@
...
+      "default": []
+    },
+    "purchasable": {
+      "type": "array",
+      "items": {
+        "$ref": "https://github.com/meetnearme/api/architecture/Purchasable.json"
+      },
```
brianfeister, line 56: Added this because we will now be supporting event tickets in v1.

architecture/purchasable.schema.json
```diff
@@ -0,0 +1,52 @@
+{
```
brianfeister, line 1: Added this because we will now be supporting event tickets in v1, tried to make this flexible to accomodate a variety of known use cases (The Regular handles memberships for example) while also an array of purchaseable for different ticket "tiers", e.g.

```
event : {
   title: "Taylor Swift at Madison Square Garden",
   purchasable: [{
     name: "VIP Section",
     cost: 300,
     currency: "USD",
     inventory: 100,
   },{
     name: "General Admission",
     cost: 100,
     currency: "USD",
     inventory: 50,000,
   }],
   ...
}

File: architecture/category.schema.json [hunk 17]
```diff
@@ -0,0 +1,24 @@
...
+      "items": {
+        "type": "string"
+      },
+      "default": []
+    },
+    "createdAt": {
+      "$ref": "https://github.com/meetnearme/api/architecture/Shared.json#definitions/ISOString"
+    }
+  },
+  "required": ["id", "name", "createdAt"]
```
brianfeister, line 23: Why is `createdAt` a required key for `category` type? I wasn't envisioning much change in this area

architecture/event.schema.json
```diff
@@ -0,0 +1,128 @@
...
+    },
+    "endTime": {
+      "$ref": "https://github.com/meetnearme/api/architecture/Shared.json#definitions/ISOString"
+    },
+    "eventCapacity": {
+      "type": "number"
+    },
+    "eventCost": {
+      "type": "number"
+    },
```
brianfeister, line 32: Going to drop these, as they're replaced in `purchasable.schema.json`

architecture/user.schema.json
```diff
@@ -0,0 +1,134 @@
...
+    "createdAt": {
+      "$ref": "https://github.com/meetnearme/api/architecture/Shared.json#definitions/ISOString"
+    },
+    "lastLoggedIn": {
+      "$ref": "https://github.com/meetnearme/api/architecture/Shared.json#definitions/ISOString"
+    },
+    "favoriteCategories": {
+      "type": "array",
+      "items": {
+        "$ref": "https://github.com/meetnearme/api/architecture/Category.json"
```
brianfeister, line 92: This should be an `items` array of UUID's shouldn't it?

architecture/user.schema.json
```diff
@@ -0,0 +1,134 @@
...
+      },
+      "default": []
+    },
+    "favoriteCurators": {
+      "type": "array",
+      "items": {
+        "$ref": "https://github.com/meetnearme/api/architecture/User.json"
```
brianfeister, line 99: This should be an `items` array of UUID's shouldn't it?

architecture/user.schema.json
```diff
@@ -0,0 +1,134 @@
...
+      },
+      "default": []
+    },
+    "mutedCurators": {
+      "type": "array",
+      "items": {
+        "$ref": "https://github.com/meetnearme/api/architecture/User.json"
```
brianfeister, line 106: This should be an `items` array of UUID's shouldn't it?

architecture/user.schema.json
```diff
@@ -0,0 +1,134 @@
...
+      },
+      "default": []
+    },
+    "eventsHosted": {
+      "type": "array",
+      "items": {
+        "$ref": "https://github.com/meetnearme/api/architecture/Event.json"
```
brianfeister, line 113: This should be an `items` array of UUID's shouldn't it?

go.mod
```diff
@@ -18,6 +18,8 @@ require (
 	github.com/go-playground/validator v9.31.0+incompatible
 	github.com/google/uuid v1.6.0
 	github.com/gorilla/mux v1.8.1
+	github.com/zitadel/oidc/v3 v3.25.1
+	github.com/zitadel/zitadel-go/v3 v3.0.0-next.2
 )
```
brianfeister, line 23: did you run `go mod tidy`? Why are we changing dependencies in what should be a JSON-only add-only architecture proposal?

architecture/event.schema.json
```diff
@@ -0,0 +1,122 @@
...
+    "recurrenceRule",
+    "recurrenceId",
+    "url",
+    "urlDomain",
+    "urlPath",
+    "urlSlug",
+    "locationLatitude",
+    "locationLongitude",
+    "locationAddress",
+    "locationCountry",
```
Brandon-G-Tripp, line 115: I believe we also need the zOrderIndex as a field on the schema.

architecture/eventRSVP.schema.json
Brandon-G-Tripp: We may want that data later, though, for the karma structure correct? That way they can get points for that and then change it to more for actual attendance. Even though that is further in the future.

architecture/user.schema.json
Brandon-G-Tripp, line 1: Here were some ideas of possible fields we are missing. Likely would not use all of these at the beginning but later on we may need it built into the schema already 

"capacity": Integer for limited capacity events
"timezone": String, critical for international events
"accessibilityFeatures": Array of strings (e.g., ["wheelchair_accessible", "sign_language_interpreter"])
 "status": String enum (e.g., "scheduled", "cancelled", "postponed")  
"isFeatured": Boolean, if you want to highlight certain events
 "attendeeCount": Integer, for displaying popularity
 "visibility": String enum ("public", "private", "unlisted")
"doorsOpenTime": ISOString
"socialMediaLinks": Object
"minimumAge" and "maximumAge": Integers
"lastAnnouncement" and "lastAnnouncementDate"
"priceRange": Object with "min" and "max" fields

architecture/eventAttended.schema.json
Brandon-G-Tripp, line 1: Possible fields:

"attendanceStatus": String enum (e.g., "confirmed", "checked-in", "no-show")
"checkInTime": ISOString (When the user checked in to the event)
"attendanceProof": String (e.g., QR code or ticket number)
"companion": Number (Number of additional people attending with the user) - I imagine this could be useful for capacity check since we will be doing ticketing of some sort  very quickly.

architecture/eventSaved.schema.json
Brandon-G-Tripp, line 1: Possible fields - 

"reminderSet": Boolean (If the user has set a reminder for this event) - this could be a check for the calendar being set or allowed, may not need it. 

"saveType": String enum (e.g., "interested", "going", "maybe")
"sharedWith": Array of strings (User IDs of friends the event was shared with) - whether this is internal or external it seems to be useful for karma and tracking for marketing purposes.

architecture/eventRSVP.schema.json
brianfeister: Yeah, I suppose we can just set this to `yes` for now and leave the other two for the future. Also, my idea of a very "lean" MVP is a bit out the window with the robust features Brandolon is looking for. Changing RSVP status is probably "out" for initial release, but will come in some time not so long after

architecture/user.schema.json
brianfeister: > "capacity": Integer for limited capacity events

`capacity` is called "`inventory`" so that it simply limits the number of any available `purchasable` item. In the case of things like "volunteer" events of "free ticketed" events, it's possible to set the cost to `0` so that the platform is just tracking registration (e.g. `capacity`)

https://github.com/meetnearme/api/pull/62/files#r1685852137

brianfeister: > "timezone": String, critical for international events

You're commenting on the `user.schema.json` but I think this is intended for `events.schema.json`?

I'm a bit on the fence about timezones, I think we can store everything in UTC on the server and then 

Especially challenging with setting a timezone as part of the `event` schema is the fact that `seshu` will be pulling these from places without sufficient data to properly derive timezones. We do need to derive UTC for `seshu` to have the (required) `event.startTime` data, but I'm not confident `event.timezone` (would need to be required if we're depending on it) is the right direction because we know we have `lat` / `lon`, which are difficult to guarantee, but necessary and we're prepared to drop events that can't provide it. Adding an extra "required" derived value that the scraping service struggles to get right (and thus high likelihood of dropping non-conforming events) seems problematic

Maybe @msnorth has ideas?

brianfeister: >"accessibilityFeatures": Array of strings (e.g., ["wheelchair_accessible", "sign_language_interpreter"])
Good call

> "status": String enum (e.g., "scheduled", "cancelled", "postponed")

I think I mentioned this elsewhere, but yes. Brandolon even wants to do something like "proposed" to facilitate venue owners / hosts discovering the interest level for "proposed" (theoretical) events, but I'm pretty uncomfortable with this from a "muck up the UI / UX with complexity perspective"

> "isFeatured": Boolean, if you want to highlight certain events

Yeah, and we might even want an `enum` for different contexts that featuring might happen. Essentially a priority signal for "feature in 'related' events on event details page" compared to "feature with a special row color in home page list"

> "attendeeCount": Integer, for displaying popularity

Should be derived from aggregate inventory of `purchasable[].itemType === 'ticket'` as demonstrated here https://github.com/meetnearme/api/pull/62/files#r1685852137

> "visibility": String enum ("public", "private", "unlisted")

Right on 💪 forgot about this one, should definitely happen for those "super secret follower-only invite" swanky vibes 

> "doorsOpenTime": ISOString

I'm looking to kill all the complexity we can. I'm on the fence about how important this is. Can't people look at their Ticketmaster.com tickets to find this? lol

> "socialMediaLinks": Object

Hmmm I think this dovetails with my mention of `externalUrl` here, we may want this to be an array and we may want it to just be an object array with a `name` key (e.g. Twitter, Facebook) and a `url` key (URL format) so that the UI can act as a arbitrary builder in case people have some weird platform that is their go-to for listing / details

> "minimumAge" and "maximumAge": Integers

Yeah, good call

> "lastAnnouncement" and "lastAnnouncementDate"

What is this? I'm totally stumped 🤔 

> "priceRange": Object with "min" and "max" fields

I think we derive this via `purchasable` and a calculation that calculates highest / lowest cost amongst `purchasable` although we may want to restrict that to a scope of only `itemType = 'ticket'`

architecture/eventAttended.schema.json
brianfeister: > "attendanceStatus": String enum (e.g., "confirmed", "checked-in", "no-show")

Hmmm, on the fence on this one, though I'll admit it aligns with **Brandolon** (github user to-be-added)

> "checkInTime": ISOString (When the user checked in to the event)

Hmmm, don't think we need this do we? Only useful if we're "measuring the user's punctuality" which feels invasive to me

> "attendanceProof": String (e.g., QR code or ticket number)

Yeah, maybe if we're doing QR code scans at the door, but as always, I opt for simplicity where possible. But **Brandolon** has a goal of event analytics that might mean we need to care about the difference between "signaled interest via RSVP" and "actually attended"

> "companion": Number (Number of additional people attending with the user) - I imagine this could be useful for capacity check since we will be doing ticketing of some sort very quickly.

This should be derived as part of `registrationFields` (allowing event host to define custom fields that allow the entering of multiple guests from the same purchase) and `purchasable` (one user can purchase multiple tickets, with multiple different guest names, but those are distinct as `purchasable` items with different quantities... and the aggregate capacity / availability is the aggregate of those purchasable items, which also includes the concept of "ticket tiers" like "VIP")

architecture/eventSaved.schema.json
brianfeister: > "reminderSet": Boolean (If the user has set a reminder for this event) - this could be a check for the calendar being set or allowed, may not need it.

Hoping to use people's real calendars (which has it's own reminder functionality) to "Add to Calendar" from the UI

> "saveType": String enum (e.g., "interested", "going", "maybe")

I might be a bit slow to comprehend here, but how is this not duplicative of this conversation https://github.com/meetnearme/api/pull/62#discussion_r1684398350

> "sharedWith": Array of strings (User IDs of friends the event was shared with) - whether this is internal or external it seems to be useful for karma and tracking for marketing purposes.

I'd rather just have a "share link" that gives you karma for someone clicking it, instead of actually tracking the people who have RSVPd after you invited them. But I imagine the analytics side will want to have a real graph. But also, I just want to caution us here as I don't believe moving toward a graph database is the correct direction at this time, so the simpler option might be better.

architecture/event.schema.json
brianfeister: Yeah, definitely needs to be added

architecture/purchasable.schema.json
brianfeister: 🚨 This is super important -- pretty sure that the `events[].purchasable` should be a list of UUIDs that point to a separate `PURCHASABLE` table, mainly because I realize we're going to want to query on `purchasable.cost` or queries like `SELECT ALL FROM PURCHASABLE WHERE itemType = 'ticket' AND inventory > 500` (sorry, I barely know SQL). 

This will play a key role in enabling the query of event sales / analytics on them, and doing table joins with the `ITEMS_SOLD` table (useful for event analytics as we move toward being "the best events management platform")

The question I have, is whether I can store an array of IDs (here in `events` table which is Dynamo) that point to the `PURCHASABLE` table

architecture/event.schema.json
brianfeister: @ItsBrandoB something to think about here

architecture/event.schema.json
brianfeister: @ItsBrandoB something to think about here

architecture/eventAttended.schema.json
brianfeister: @ItsBrandoB something to think about here

architecture/eventRSVP.schema.json
brianfeister: @ItsBrandoB something to think about here

architecture/eventSaved.schema.json
brianfeister: @ItsBrandoB something to think about here

architecture/purchasable.schema.json
brianfeister: @ItsBrandoB something to think about here

architecture/user.schema.json
brianfeister: @ItsBrandoB something to think about here

architecture/user.schema.json
```diff
@@ -0,0 +1,134 @@
...
+    }
+  },
+  "required": [
+    "id",
+    "firstName",
+    "lastName",
+    "karma",
+    "email",
+    "isEmailVerified",
+    "dateOfBirth",
```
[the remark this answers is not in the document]
brianfeister: @ItsBrandoB something to think about here

architecture/category.schema.json
j4inam: Yeah we need to have it exist on github first

architecture/user.schema.json
j4inam: Makes sense. We can configure custom perms on zitadel

architecture/eventRSVP.schema.json
j4inam: My motive here was:

Say I am having some closed event like a babyshower. If I invite you, I'd assume you're attending unless you explicitly decline.

architecture/category.schema.json
j4inam: I'd want to have createdAt information for all entities we have in the db. May not need `updatedAt` but I do want `createdAt` for everything.

architecture/user.schema.json
j4inam: Yeah. I left the reference in for all join-like properties to show what does that UUID reference ?

architecture/user.schema.json
j4inam: Yeah. I left the reference in for all join-like properties to show what does that UUID reference ?

architecture/user.schema.json
j4inam: Yeah. I left the reference in for all join-like properties to show what does that UUID reference ?

architecture/user.schema.json
j4inam: Yeah. I left the reference in for all join-like properties to show what does that UUID reference ?

architecture/user.schema.json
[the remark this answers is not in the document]
j4inam: We can skip the year of birth. Just ask for month and day.

go.mod
j4inam: I didn't really touch dependencies.

architecture/purchasable.schema.json
```diff
@@ -0,0 +1,68 @@
...
+  "$id": "https://github.com/meetnearme/api/architecture/Purchasable.json",
```
brianfeister, line 2: TODO: we need to add a `Purchases` (includes $0 transactions, so could be a different name) DB table for storing `registrationFields` responses, `Stripe transaction ID`, `items`, `total`, etc.

architecture/user.schema.json
```diff
@@ -0,0 +1,128 @@
+{
```
brianfeister, line 1: TODO: add "reshares" to the user schema, facilitating cross-promotion of events 

Add "muted" users 

Add concept of organization ( even if it's one user per organization to start and we don't really implement it at the auth layer yet)

architecture/user.schema.json
brianfeister: I'm thinking about this some more @itsbrandoB, for analytics, it would be really useful to have the users Birth Month / Year. Marketing would love to have their birthdate I'm sure. But the demographic analytics unlocked with their age is crazy valuable, though I wish it didn't create the privacy-intruding dynamic. 

Thoughts @j4inam @msnorth @Brandon-G-Tripp ?

architecture/category.schema.json
brianfeister: My hope was for some relative linking. Because it'll be bad for this to always point to "current `master`" so that changes fail JSON schema validation due to branch reference.

architecture/category.schema.json
```diff
@@ -0,0 +1,24 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "id": {
+      "$ref": "https://github.com/meetnearme/api/architecture/Shared.json#definitions/UUID"
+    },
+    "name": {
+      "type": "string"
+    },
+    "subCategories": {
```
brianfeister, line 12: Missing `id` on subcategory

architecture/event.schema.json
brianfeister: This is resolved by the key `isExternalEvent` here

https://github.com/meetnearme/api/pull/62/files#diff-1a51ab75678fa60b0271a1522c7d31b1625d6e574a3dc90dccd6bfe57ac99ebbR86

architecture/event.schema.json
brianfeister: TODO: implement the example above and add to the json schema file here

brianfeister: DECISION: these are our initial types

* Email
* URL
* Text Input
* Check box
* Multi Select

architecture/event.schema.json
```diff
@@ -0,0 +1,126 @@
...
+    },
+    "urlPath": {
+      "type": "string"
+    },
+    "urlQueryParams": {
+      "type": "string"
+    },
+    "urlSlug": {
+      "type": "string"
+    },
```
brianfeister, line 59: TODO: delete this

architecture/event.schema.json
```diff
@@ -0,0 +1,126 @@
...
+    "url",
+    "urlDomain",
+    "urlPath",
+    "urlSlug",
+    "locationLatitude",
+    "locationLongitude",
+    "locationAddress",
+    "locationCountry",
+    "locationZipcode",
+    "isVerified",
```
brianfeister, line 120: DELETE

architecture/eventSaved.schema.json
```diff
@@ -0,0 +1,22 @@
+{
```
brianfeister, line 1: DECISION: drop this, we want people to put `maybe` if all they're doing is thinking about it

architecture/eventRSVP.schema.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$id": "https://github.com/meetnearme/api/architecture/EventRSVP.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "definitions": {
+    "RSVPStatus": {
+      "type": "string",
+      "enum": ["yes", "no", "maybe"]
```
brianfeister, line 7: DECISION: Add `Interested, can't make it` to this enum

DECISION: Drop `No` here. We're not doing an "evite.com" style event

msnorth (approved):
lgtm

architecture/eventAttended.schema.json
brianfeister: I've deleted this one because "confirmation of actual attendance" is fine for roadmap goals, but not on our near-term radar. Thus it should be revisited in the future with more information

architecture/user.schema.json
brianfeister: DECISION: deleting this, will live in Zitadel

architecture/category.schema.json
brianfeister: Seems maybe impossible to do what I want here

Brandon-G-Tripp (approved):
Looks good after our discussion today.

Brandon-G-Tripp:
@brianfeister I know I approved but then saw the changes to the go.mod that you had asked Jainam about. Do you have a resolution idea for those?

PR #64, opened 2024-07-22
Attempting delete stacks/DeployProd.ts to check if anything breaks

brianfeister:
This file appears not to be used, I am removing it as a test. **HOWEVER**, it's also possible that solving our DB schema changes that break deployments involve using CDK more directly and this could be useful in that direction. 

This file was added as part of following the SST docs here:

https://docs.sst.dev/going-to-production#stacks-setup


brianfeister:
YOLO merging because the only way to test this is to see what happens (via Github Actions) when I merge the PR

PR #67, opened 2024-07-25
Merge Back `develop` => `main`

brianfeister:
I'm going to self-merge this first one, only because it simply adds the `TODO:` search github actions check that I want us to all standardize on as we will be moving starting quite soon


functions/gateway/helpers/db_helpers.go
```diff
@@ -13,7 +13,7 @@ func IsRemoteDB() bool {
     }
 	sstStage := os.Getenv("SST_STAGE")
 	// `.github/workflows/deploy-feature.yml` deploys any branch that begins with `feature/*` to aws as `feature-*`
-	return sstStage == "prod" || strings.HasPrefix(sstStage, "feature-")
```
brianfeister, line 16: This was a change I had to make on `develop` because the `sstStage` was NOT `prod` and the `feature-` prefix (only other condition for this remote DB helper to point remote) was absent. This allowed `develop` to deploy

PR #81, opened 2024-07-26
Update go.yml for Go tests

Brandon-G-Tripp:
Add run of Go tests to workflow file and coverage report. Uploads coverage report to repository for reference.

Patch to add CI check for test coverage at 50% and all tests passing.


brianfeister (changes requested):
It looks like there's an error in the grep command and the coverage check you added is failing, but not due to threshold

brianfeister:
Also, do we want to be piping these to txt files? I'm not sure what the exact limit is, but GitHub allows for artifacts from test runs but storage is not unlimited

brianfeister:
Sorry, still processing this. Sooooo is exiting `1` expected behavior because that signals that the threshold isn't met? If that's the case I think we want to echo something in the output rather than silently failing without explanation that the coverage threshold is unmet

.github/workflows/go.yml
```diff
@@ -32,3 +33,29 @@ jobs:
...
+
+      - name: Check coverage
+        run: |
+          COVERAGE=$(awk -F'[[:space:]]+' '/^total:/ {print $NF+0}' coverage.txt)
+          echo "Total coverage: $COVERAGE%"
+          if (( $(echo "$COVERAGE < 50" | bc -l) )); then
+            echo "🔴 Coverage is $COVERAGE%. We require minimum 50%."  
+            exit 1
+          elif (( $(echo "$COVERAGE < 60" | bc -l) )); then
+            echo "🟠 Coverage is $COVERAGE%. It's passing but is close to the failure cutoff."
```
msnorth, line 51: I think Java had some sort of utility we used which made sure our coverage never decreased. This is good for now but something to think of

msnorth (approved):

brianfeister (approved):

brianfeister (approved):

PR #85, opened 2024-07-30
Generate _templ.go files in git workflow

msnorth:
This change adds the _templ.go files to the gitignore
and adds their generation to the go workflow step.

The purpose of this change is to prevent generated
files from cluttering up our code and ensure deterministic
prod artifact generation.


Brandon-G-Tripp:
The failure is interesting because it is now trying to fetch our internal packages with the go get command. I think that is why it is failing but running in locally I see no issues with it at all. @bfeister did you run into any issues with the go get ./... in your previous work with the workflow files?

brianfeister:
Attempting this @msnorth @Brandon-G-Tripp 

https://github.com/meetnearme/api/pull/85/commits/7e9527059e7393f10ac58224d83e770e1d163d71

brianfeister:
Oh I see the problem. You can't delete the `*_templ.go` files because other packages reference the Go module `partials` and `pages`, which are not present when you delete the module declaration here

https://github.com/meetnearme/api/pull/85/files#diff-51ff350d65aa855a1de36343f97f9eab0f5a1ddad4947bbfe736da8a417d584dL4

The risk of out of sync `*_templ.go` files might need a different approach

Brandon-G-Tripp:
@brianfeister Yeah I saw the same issue. We are missing the built dependencies. We can just run the templ generate before go get correct? So it generates all of the files for reference then go get to get the external dependencies and also build the dependency tree I assume it builds under the hood.

brianfeister:
@Brandon-G-Tripp 

> We can just run the templ generate before go get correct? 

Smart idea 💡  ... just commit on this branch and test that theory

Brandon-G-Tripp:
@brianfeister It works in the PR i opened against this gitignore PR. 

https://github.com/meetnearme/api/pull/86

brianfeister:
LGTM :shipit:

brianfeister (approved):
🔥

brianfeister:
Whoops, we broke things because `go.yml` runs in a separate runner from `Deploy SST <Dev | Prod> `

PR #86, opened 2024-07-30
Update go.yml

Brandon-G-Tripp:
We must generate _templ.go files before we allow Go to deal with external and internal dependencies with the go get command.


msnorth (approved):
shipit

Issue #65, opened 2024-07-22
[bug] [CI / SST] Failing deployments to prod


brianfeister:
Right now, deployments are failing to `prod` with the error

```
Errors
   StorageStack UPDATE_ROLLBACK_COMPLETE
   SeshuSessions/Table: Resource handler returned message: "Resource of type 'AWS::DynamoDB::Table' with identifier 'prod-meetnearme-go-fullstack-SeshuSessions' already exists." (RequestToken: f66d61f8-926d-2398-f5b9-5fba6de15c54, HandlerErrorCode: AlreadyExists)
```

As alluded to in #64 , it seems there's something we need to do in a more granular way (perhaps using AWS CDK directly) to deploy schema changes to the DynamoDB table without triggering a failure via attempt to deploy the table as an entirely new resource

PR #87, opened 2024-07-31
Merge back: `main` => `develop`

brianfeister:
This one's non-controversial and I will likely self-merge. It's just the #86 and #85 changes to drop `templ` compiled Go files


brianfeister:
Only failure is `SST Deploy Prod` which is known to be broken re: #65

PR #90, opened 2024-08-01
Merge back: `develop` => `main`

brianfeister:
Prior to landing the last (currently) open PR for the Seshu Handoff https://github.com/meetnearme/api/pull/60, we want to promote `develop` to `main`. The seshu handoff PR is intended to be deployed / promoted to `develop` as part of testing and continued work, but we don't want all of those open `TODO:`s in `main`.

Hence, we want to merge back now.


functions/gateway/handlers/page_handlers.go
```diff
@@ -13,20 +14,19 @@ import (
...
-var mw   *authentication.Interceptor[*openid.UserInfoContext[*oidc.IDTokenClaims, *oidc.UserInfo]]
+var mw *authentication.Interceptor[*openid.UserInfoContext[*oidc.IDTokenClaims, *oidc.UserInfo]]
 
-func init () {
-	Db = transport.GetDB()
-	mw, _ = services.GetAuthMw()
+var bypassAuthForTesting bool
+
+func SetBypassAuthForTesting(bypass bool) {
+    bypassAuthForTesting = bypass
```
Brandon-G-Tripp, line 29: I will double check but I believe we can remove as was never fully implemented.

brianfeister: Alright, I'll look for a new PR to `develop` to fix this before we promote to `main`

Brandon-G-Tripp (commented):
This looks good. I am just going to remove the small unused code that I had flagged.

brianfeister:
> This looks good. I am just going to remove the small unused code that I had flagged.

@Brandon-G-Tripp open a PR aimed at `develop` if you're fixing that. When it merges to `develop` it automatically will appear in the commit change set in this PR.

You don't want to commit it directly to `develop` (which is how it would wind up in this PR)

Brandon-G-Tripp (approved):
Ok the clean up is finished. This is good to promote to main.

PR #91, opened 2024-08-01
Prune legacy node deps

brianfeister:
It's (maybe) cool (sometimes definitely) that dependabot scans our repo automatically. 

This scan https://github.com/meetnearme/api/security/dependabot made me realize that we have a bunch of node deps from an old POC I did that is not used anywhere in code


brianfeister:
Self-merging. This is non-controversial

PR #113, opened 2024-08-12
[Feature] New Theme + Early Geolocation / Selection Iteration

brianfeister:
This PR implements a new theme (screenshots below) per the [mocks from Jeremi Dudu here](https://www.figma.com/design/6m2pazqCK7t4z6Mlbbf50q/Untitled?node-id=68-2&t=tYkxfMCvfqttkZ3C-0)

There might still be some mis-alignments in terms of theme / styling, feel free to point them out in comments here, but also some of these will evolve as the conversation between Jeremi and I evolves

In addition, I have implemented an Alpine.js + Tailwind combobox UI element (can be componentized later) that does two things:

1. Shows the implicit geolocation to the user ...we can get this via 
    1. Cloudflare Ray ID (which is suffixed with the 3 letter airport code like `DEN` (Denver) 
    1. `lat` and `lon` querystring params
1. Allows the user to search and select from 200 of the largest US cities (selecting an item does nothing, which leads to the open question I'm pondering)
1. Add `/events/{event id}` (arguably this should be `/event` from a REST verb perspective) event detail page with a (temporarily) `O(n)` query until we get the GSI for `event.id` in place

**Open Question** – As always, I'm hyper-paranoid about keeping cost-to-serve low enough that 1 paying user can cover the infra expenses of 1,000+ non-paying users. As such, I'm unsure if we should use the Google Geolocation API in the search bar or just have a static list of "suggestions" for largest cities. This mainly prevents abuse / bots from getting free data out of Google Maps Geolocation API (where we would be paying for / targeted by malicious attackers)

🤔 As I ponder this some more, I'm thinking a list of `200` biggest cities I send down in the HTML response as core suggestions, but then a static / simple string based list on a backend API of our own that has 1,000 biggest cities. That should cover things for early days without sending the full `1000` cities on every home page load

## Screenshots

image

image

image

image


.github/workflows/deploy-feature.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Generate templ code
         uses: './.github/actions/templ_generate'
         with:
-          templ-version: 'v0.2.663'
+          templ-version: 'v0.2.747'
```
brianfeister, line 34: We're using a `templ` feature from `'v0.2.747'` hence the upgrade. 

Also, `templ` [supports HTTP streaming](https://templ.guide/server-side-rendering/streaming/), which is nuts and we definitely should use for the snappiest TTFB feel

functions/gateway/services/events_service.go
```diff
@@ -117,16 +117,52 @@ func GetEventsZOrder(ctx context.Context, db internal_types.DynamoDBAPI, startTi
     return events, nil
 }
 
+
+func GetEvent(ctx context.Context, db internal_types.DynamoDBAPI, eventId string) (*EventSelect, error) {
```
Brandon-G-Tripp, line 121: @brianfeister Should we name this to be GetEventById  in case was have some other handlers for separate GSI's we could eventually generalize it to also take a flag for what GSI we are querying on so we do not need a bunch of duplicate code.

Brandon-G-Tripp (changes requested):
This looks good to me. I think the only thing missing is a little test coverage. Maybe the CFray func and adding checks for the cfLocation, rayCode, etc.

static/assets/global.css
```diff
@@ -157,3 +157,36 @@
 }
 
 /* END seshu ingestion "add event source" section */
+
+/* TODO: delete this hack, temporary for visual demo */
```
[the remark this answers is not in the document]
brianfeister: removed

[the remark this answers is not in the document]
brianfeister: removed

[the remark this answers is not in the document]
brianfeister: removed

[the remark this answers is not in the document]
brianfeister: removed

functions/gateway/handlers/page_handlers.go
```diff
@@ -59,12 +59,14 @@ func GetHomePage(w http.ResponseWriter, r *http.Request) http.HandlerFunc {
...
   cfRay := GetCfRay(ctx)
   rayCode := ""
+	var cfLocation helpers.CdnLocation
   cfLocationLat := services.InitialEmptyLatLon
   cfLocationLon := services.InitialEmptyLatLon
   if len(cfRay) > 2 {
     rayCode = cfRay[len(cfRay)-3:]
-	  cfLocationLat = helpers.CfLocationMap[rayCode].Lat
-    cfLocationLon = helpers.CfLocationMap[rayCode].Lon
+		cfLocation = helpers.CfLocationMap[rayCode]
```
Brandon-G-Tripp, line 67: This is all looking pretty good. This CfLocationMap is generated in CI but it is making the local compilation fail when running tests because it is only generating in CI. Do you have a work around for this? I am hesitant to approve this unless we have a path to successfully compile and test locally. 

I wonder if we are almost at the point of needing a Makefile or similar to manage our testing and build commands locally if we are going to start running into exceptions where we need certain code generate to test. 

@brianfeister

functions/gateway/handlers/page_handlers_test.go
```diff
@@ -142,14 +132,23 @@ func TestGetMapEmbedPage(t *testing.T) {
...
+    PathParameters: map[string]string{
+        helpers.EVENT_ID_KEY: eventID,
+    },
+	})
+	req = req.WithContext(ctx)
+
 	// Set up router to extract variables
 	router := mux.NewRouter()
-	router.HandleFunc("/events/{eventId}", func(w http.ResponseWriter, r *http.Request) {
+	router.HandleFunc("/events/{" + helpers.EVENT_ID_KEY + "}", func(w http.ResponseWriter, r *http.Request) {
```
Brandon-G-Tripp, line 151: That is great that you got the test working. Is this the key here with setting up the handleFunc along with the custom scanFunc for the mock?

Brandon-G-Tripp (changes requested):
My only comment is that potential issue with the ungenerated template for CfLocation. If there is a workaround or solution please document and then we are ready for merge. @brianfeister

functions/gateway/handlers/page_handlers.go
brianfeister: We have a path to compile locally. You just do it manually one time by copy/pasting the JSON from https://speed.cloudflare.com/locations into the gitignored `cloudflare_locations.go` file. Since it's gitignored, you'll be able to keep it around and local dev will work fine. The CF locations are **extremely** slow in change rate so this isn't a big risk, at least not for probably 12+ months from now

We could write a script to generate it, but for now, I'm comfortable with the manual step since we're needing to move fast

brianfeister: For clarity:

1. Open `cloudflare_locations_template` 
1. "Save as" `cloudflare_locations.go`
1. Copy paste https://speed.cloudflare.com/locations over the `<replace me>` string
1. Profit

brianfeister: Also, important note here: what we do in local dev is actually **irrelevant** because none of this has a way of working in local dev, we're fully reliant on lower envs like `devnear.me` and `testingnear.me` to validate logic pertaining to Cloudflare

Brandon-G-Tripp (approved):
This is good to ship. We can add a tiny PR to add procedure to READme unless I missed it.

functions/gateway/handlers/page_handlers_test.go
brianfeister: Mainly the mock needed an `if` branch statement to know whether the query was coming in via `id`. That will go away once we do the proper `db.GetItem()` call and have GSIs set up via @msnorth 's work in the DB layer

PR #115, opened 2024-08-28
Cloudflare KV Edge Lookup for Subdomains

brianfeister:
This PR is a partial / first pass, introducing 

1. An API endpoint for adding a new item to KV store from within your `/admin/profile` section
2. An appended header that is (for now) the user ID of the user who added that header. This header will be used to make "scoped" DB requests that only show "events owned by org with ID <X-Mnm-Subdomain-Value>"

### Testing this

1. Go to https://testingnear.me/admin/profile (has to be a test env that has cloudflare in front of it)
1. Add a value in the `Set Subdomain` input field
1. Hit `Update Subdomain`
1. Open a new browser tab
1. Open Dev Tools
1. Go to `<subdomain>.testingnear.me`
1. Look for the main request for that page and verify that there is a header `X-Mnm-Subdomain-Value` and that it's value is the same as what you see for `ID` under "My Info" on `/admin/profile`
image

### Problems

~⚠️ Users can currently "grab" multiple subdomains and we need logic to prevent abuse / greedy subdomain grabbing~ (this is fixed now, see screenshot)

https://github.com/user-attachments/assets/b14e1b5f-7de2-495e-ac34-0370035fdfad

image


.editorconfig
Brandon-G-Tripp, line 1: Is this meant to be checked in for maintaining consistent formatting across the team?

.env.example
Brandon-G-Tripp, line 1: @brianfeister can we get these new env variables documented in the README.

.github/actions/set_aws_creds_env_vars/action.yml
Brandon-G-Tripp, line 1: I see that the cloudflare values are used as inputs but I do not see them used in the action. Are they just used internally for the worker?

worker/index.js
```diff
@@ -0,0 +1,41 @@
...
+    const hostname = url.hostname;
+    const parts = hostname.split('.');
+    const subdomain = parts.slice(0, -2).join('.');
+
+    console.log('req.cf: ', JSON.stringify(req.cf));
+
+    let subdomainValue = null;
+
+    // Clone the original request
+    let newRequest = new Request(req);
```
Brandon-G-Tripp, line 13: I am getting a linter warning to have this be const since it is never reassigned.

Brandon-G-Tripp (changes requested):
I just have the couple of small comments to change and add documentation of new env variables and then a bit more test coverage. Should we force past the coverage for now?

worker/index.js
brianfeister: Will fix

.github/actions/set_aws_creds_env_vars/action.yml
brianfeister: They're used over here in the Cloudflare / Wrangler worker deploy

https://github.com/meetnearme/api/pull/115/files/4324f6f90f2bb3d71d3f662812fd4de9c9dccc1e#diff-b2da012ee27f5cf144e34765250f8d30034a50f4073447502c81cddb8c60178fR67-R71

brianfeister: Also used here in the route handler

https://github.com/meetnearme/api/pull/115/files/4324f6f90f2bb3d71d3f662812fd4de9c9dccc1e#diff-019ebe66f31368757e15573c1255b7a2e0403b52bf1e6170e6833ee64f491e10R56-R57

.editorconfig
brianfeister: Yeah, this is to help me (and others), seems 2 tab indent is the go default via `gofmt` which we need to introduce for CI linting checks

brianfeister:
I'm hoping to get the coverage up and stick to our promise to ourselves. Admittedly, it's a bit tough with our tight timeline

.env.example
brianfeister: Will do

brianfeister:
Coverage resolved @Brandon-G-Tripp mocking that new controller I just added for the KV work. LFG!!! 🔥  Let's merge this before things get crazier. 

Cursor.ai did insane heavy lifting getting me through writing that test... was quite a big chunk of code. 💪

.env.example
brianfeister: Actually, I'm not expecting us to document env vars. `.env.example` is the documentation and **must** be looked at. I'll add a note to look there. I've updated this

Brandon-G-Tripp, line 1: Beautiful! Thank you. This will save us when we forget all of our variables.

Brandon-G-Tripp (approved):
Love to see this!

PR #122, opened 2024-09-03
User Auth & Signup

j4inam:
## Description

Implement authentication for users and middleware.

## Changes
- Redo User Login to implement PKCE flow

https://github.com/user-attachments/assets/2a3b4257-244b-48ed-b41e-d6a46f4c6b9d


brianfeister:
@j4inam your builds have been failing all along, I think you might have a missing environment variable in GitHub. We must have a successful deploy before we can merge

functions/gateway/services/auth_service.go
```diff
@@ -2,55 +2,187 @@ package services
...
-	key         = flag.String("key", os.Getenv("ZITADEL_ENCRYPTION_KEY"), "encryption key")
-	clientID    = flag.String("clientID", os.Getenv("ZITADEL_CLIENT_ID"), "clientID provided by ZITADEL")
-	redirectURI = flag.String("redirectURI", string( os.Getenv("APEX_URL") + "/auth/callback"), "redirect URI registered with ZITADEL")
-	mw   *authentication.Interceptor[*openid.UserInfoContext[*oidc.IDTokenClaims, *oidc.UserInfo]]
-	authN *authentication.Authenticator[*openid.UserInfoContext[*oidc.IDTokenClaims, *oidc.UserInfo]]
-	once sync.Once
+	domain        = flag.String("domain", os.Getenv("ZITADEL_INSTANCE_URL"), "your ZITADEL instance domain (in the form: https://<instance>.zitadel.cloud or https://<yourdomain>)")
+	key           = flag.String("key", os.Getenv("ZITADEL_ENCRYPTION_KEY"), "encryption key")
+	clientID      = flag.String("clientID", os.Getenv("ZITADEL_CLIENT_ID"), "clientID provided by ZITADEL")
+	clientSecret  = flag.String("clientSecret", os.Getenv("ZITADEL_CLIENT_SECRET"), "clientSecret provided by ZITADEL")
```
brianfeister, line 28: This value isn't in the organization secrets, which is likely the reason the SST builds have been failing since the beginning with your first deployments 
Screenshot_20240915-141533

j4inam: Fixed env vars

brianfeister:
@j4inam thanks for pushing so hard on this 👏 ... I know it's been late night effort, I don't want us to all die on this death march, but I'm also grateful for the hard work.

Here is what I'm looking at:

- [x] Set up API mappings in AWS (we need to get permissions figured out so you guys can do this yourselves, but you can't just now) and pointed https://john-does-pub.testingnear.me/?lat=39.8616981506&lon=-104.672996521&radius=1500 and https://brians-pub.testingnear.me/?lat=39.8616981506&lon=-104.672996521&radius=500 to your feature branch deployment to verify that the Cloudflare layer is still working (I saw your commits in that area). Good news, things look like they're working ✅ 
- [ ] Validate that nothing broke functionally... Looks like the dropdown for location selection is broken now (the Alpine.js directives got somehow messed up. I'm going to fix this now so I can hopefully merge this right after:
image

functions/gateway/main.go
```diff
@@ -54,103 +54,174 @@ func init() {
...
+				if refreshAccessTokenErr != nil {
+					log.Printf("Authentication Failed: %v", err)
+					http.Error(w, "Authentication failed", http.StatusUnauthorized)
+					return
+				}
+
+				// Store the access token and refresh token securely
+				accessToken, ok := tokens["access_token"].(string)
+				if !ok {
+					http.Error(w, "Failed to get access token", http.StatusInternalServerError)
```
brianfeister, line 130: Is there a better UX here where a customer isn't at risk for seeing an empty page with `"Failed to get access token"? This terminology won't make sense to most people. We could log it internally and instead redirect to a UI page that says "We couldn't log you in... try again?" or something like that

functions/gateway/main.go
```diff
@@ -54,103 +54,174 @@ func init() {
...
+					return
+				}
+
+				refreshToken, ok := tokens["refresh_token"].(string)
+				if !ok {
+					fmt.Printf("Refresh token error: %v", ok)
+					http.Error(w, "Failed to get refresh token", http.StatusInternalServerError)
```
brianfeister, line 137: Same as my comment above

functions/gateway/main.go
brianfeister: Probably we just want to throw a ticket on the backlog for making this more friendly

functions/gateway/main.go
```diff
@@ -54,103 +54,174 @@ func init() {
...
+	var handler http.HandlerFunc
+	var accessTokenCookie *http.Cookie
+	var refreshTokenCookie *http.Cookie
+	var err error
+	var refreshTokenCookieErr error
+	switch route.Auth {
+	case Require:
+		handler = func(w http.ResponseWriter, r *http.Request) {
+			accessTokenCookie, err = r.Cookie("access_token")
+			if err != nil {
```
brianfeister, line 113: Avoid indentation hell
```suggestion
			if err != nil {
			    return 
			}
```

functions/gateway/main.go
```diff
@@ -54,103 +54,174 @@ func init() {
...
+					Name:  "access_token",
+					Value: accessToken,
+					Path:  "/",
+				})
+
+				http.SetCookie(w, &http.Cookie{
+					Name:  "refresh_token",
+					Value: refreshToken,
+					Path:  "/",
+				})
```
brianfeister, line 152: Does zitadel have a concept of TTL? Or do we just assume "infinite" life and let zitadel decide when it's expired?

functions/gateway/main.go
```diff
@@ -54,103 +54,174 @@ func init() {
...
-}
+	var handler http.HandlerFunc
+	var accessTokenCookie *http.Cookie
+	var refreshTokenCookie *http.Cookie
+	var err error
+	var refreshTokenCookieErr error
+	switch route.Auth {
+	case Require:
+		handler = func(w http.ResponseWriter, r *http.Request) {
+			accessTokenCookie, err = r.Cookie("access_token")
```
brianfeister, line 112: Why do we check for `access_token` here? Because we assume a cookie might expire and delete itself, while the access token is long-lived?

functions/gateway/main.go
```diff
@@ -54,103 +54,174 @@ func init() {
...
+				http.Redirect(w, r, userRedirectURL, http.StatusFound)
+				return
+			}
+
+			accessToken := "Bearer " + accessTokenCookie.Value
+
+			// Use the Authorizer to introspect the access token
+			authCtx, err := app.AuthZ.CheckAuthorization(r.Context(), accessToken)
+			if err != nil {
+				http.Error(w, "Unauthorized: Invalid access token", http.StatusUnauthorized)
```
brianfeister, line 164: This is another one that should have a user-friendly error page 

"Sorry, you must be logged in to access this page" with an HTML page that includes the MNM app

functions/gateway/main.go
```diff
@@ -54,103 +54,174 @@ func init() {
...
+				return
+			}
+
+			userInfo := helpers.UserInfo{}
+			data, err := json.MarshalIndent(authCtx, "", "	")
+			if err != nil {
+				http.Error(w, "Unauthorized: Unable to fetch user information", http.StatusUnauthorized)
```
brianfeister, line 171: If we're in the `Check` mode, we shouldn't throw any errors. We should just log the error and then redirect the user  to their destination

functions/gateway/main.go
```diff
@@ -54,103 +54,174 @@ func init() {
...
+				return
+			}
+			err = json.Unmarshal(data, &userInfo)
+			if err != nil {
+				http.Error(w, "Unauthorized: Unable to fetch user information", http.StatusUnauthorized)
```
brianfeister, line 176: Same as above

functions/gateway/main.go
```diff
@@ -54,103 +54,174 @@ func init() {
...
+
+			userInfo := helpers.UserInfo{}
+			data, err := json.MarshalIndent(authCtx, "", "	")
+			if err != nil {
+				route.Handler(w, r).ServeHTTP(w, r)
+				return
+			}
+
+			err = json.Unmarshal(data, &userInfo)
+			log.Printf("User info %v", userInfo)
```
brianfeister, line 209: This is going to make our log volume pretty high, let's only log the failure to unmarshal in the `if err != nil` block blelow

brianfeister (approved):
Lots of comments, but I'm good to merge this because time is short and I'd rather SHIP something and keep moving than get slowed down in these details

PR #123, opened 2024-09-05
Update schema to new marqo-indexed fields, move `purchasable` & `registrationFields` out of event schema

brianfeister:
I'm implementing Marqo, which means we want this data to be fast and light, focusing only on what we want to index on. Fortunately, both `purchasable` and also `registrationFields` are data calls that we can resolve async (either via HTMX on load, or via `templ` http streaming if Lambda API Gateway allows it) so we want to move those out.

Also some adjustments are made based on the following marqo index configuration:

```

{
  "type": "structured",
  "vectorNumericType": "float",
  "model": "hf/bge-large-en-v1.5",
  "normalizeEmbeddings": true,
  "textPreprocessing": {
    "splitLength": 2,
    "splitOverlap": 0,
    "splitMethod": "sentence"
  },
  "imagePreprocessing": {
    "patchMethod": null
  },
  "annParameters": {
    "spaceType": "prenormalized-angular",
    "parameters": {
      "efConstruction": 512,
      "m": 16
    }
  },
  "tensorFields": ["name_description_address"],
  "allFields": [
    {
      "name": "name_description_address",
      "type": "multimodal_combination",
      "dependentFields": {"name": 0.3, "address": 0.2, "description": 0.5}
    },
    {"name": "eventOwners", "type": "text", "features": ["filter", "lexical_search"]},
    {"name": "tags", "type": "text", "features": ["filter", "lexical_search"]},
    {"name": "categories", "type": "text", "features": ["filter", "lexical_search"]}
  ]
}
```

### DIFF
--- a/schemas/event.schema.json
+++ b/schemas/event.schema.json
@@ -1,50 +1,9 @@
 {
   "$id": "https://github.com/meetnearme/api/architecture/Event.json",
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "definitions": {
-    "registrationFields": {
-      "type": "object",
-      "properties": {
-        "name": {
-          "type": "string"
-        },
-        "type": {
-          "type": "string",
-          "enum": [
-            "text",
-            "url",
-            "email",
-            "phone",
-            "number",
-            "select",
-            "checkbox",
-            "radio"
-          ]
-        },
-        "options": {
-          "type": "array",
-          "items": {
-            "type": "string"
-          }
-        },
-        "required": {
-          "type": "boolean"
-        },
-        "default": {
-          "type": "string"
-        },
-        "placeholder": {
-          "type": "string"
-        },
-        "description": {
-          "type": "string"
-        }
-      }
-    }
-  },
   "type": "object",
   "properties": {
-    "id": {
+    "_id": {
       "$ref": "https://github.com/meetnearme/api/architecture/Shared.json#definitions/UUID"
     },
     "eventSourceId": {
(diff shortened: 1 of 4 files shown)


schemas/event.schema.json
```diff
@@ -53,6 +12,13 @@
     "eventSourceType": {
       "$ref": "https://github.com/meetnearme/api/architecture/Shared.json#definitions/EventSourceType"
     },
+    "eventOwners": {
+      "type": "array",
+      "items": {
+        "$ref": "https://github.com/meetnearme/api/architecture/Shared.json#definitions/UUID"
+      },
+      "minItems": 1
+    },
```
brianfeister, line 21: Making this an array for the "chapters" concept. You can have one owner, or many. So a "UNICEF Denver" event would have an array `[<UNICEF national org id>, <UNICEF denver chapter id>]`

brianfeister: This is our core "group" filtering logic for pages that show events particular to their group

schemas/event.schema.json
```diff
@@ -68,16 +34,11 @@
...
-      "default": []
+    "hasRegistrationFields": {
+      "type": "boolean"
     },
-    "purchasable": {
-      "type": "array",
-      "items": {
-        "$ref": "https://github.com/meetnearme/api/architecture/Purchasable.json"
-      },
-      "default": []
```
brianfeister, line 80: moved out to their own tables

schemas/event.schema.json
```diff
@@ -108,9 +69,6 @@
       },
       "default": []
     },
-    "zOrderIndex": {
-      "type": "string"
-    },
```
brianfeister, line 113: Gone because of marqo

schemas/purchasable.schema.json
```diff
@@ -30,39 +30,48 @@
   },
   "type": "object",
   "properties": {
-    "name": {
-      "type": "string"
```
brianfeister, line 34: Not gone, just moved down into `items`

Brandon-G-Tripp (approved):

Brandon-G-Tripp (approved):

PR #124, opened 2024-09-05
Feature/rds tables and api

Brandon-G-Tripp:
This is the beginning of implementing rds tables and to begin discussion. 

Please view possible curl requests for the users table draft in API_ENDPOINTS.md which begins documenting the api. Please discuss and evaluate as code comes through. I will update PR and flesh out the full documentation shortly for what the PR contains. For now User table CRUD is implemented. Please also evaluate services/migrations for all of the table creation so we can solidify the columns we desire.


brianfeister:
Marked ready just to start triggering unit test runs

brianfeister (approved):

PR #183, opened 2024-10-03
fix(timezone_display_events): fix for using formatted and localized time

Brandon-G-Tripp:
and dates.

GetTimeOrShowNone and GetDateOrShowNone become wrappers for the localizing and formatted time.


brianfeister:
Fixed in #210 thanks @Brandon-G-Tripp 👍

PR #203, opened 2024-10-25
Feature/Checkout inventory rollback

### LINKED ISSUE #201: Stripe Webhook Listener that listens for Checkout success / failure / expiration and handles Purchasable Qty
For launch, we left a critical infrastructure component unfinished:

When a user clicks `Checkout` in the sidebar, we create a Stripe Checkout session for them, but behind the scenes, we optimistically assume that their transaction is successful, add it to the `purchases` table in Dynamo, and mark it with `status: PENDING`

The problem is that as part of this optimistic update, we also decrement `purchasables[<event id>][<item id>].inventory` (so if it was at `50`, it will be at `49` after a user clicks "Checkout")

The missing piece is a Stripe webhook that will push to Meet Near Me regarding the status of each Checkout sessions. For ones that expire without successful payment or otherwise error / fail, we need to listen for that webhook event and increment / restore the `purchasables[<event id>][<item id>].inventory` for that item

At a high level, this means that in a high ticket volume scenario, the first person to click "Checkout" has effectively "reserved" their ticket for the duration defined by the checkout session (currently 30 min). 

The effect for users is that this is possible: 

1. [6:00pm] `user1` "reserves" the last ticket for 30 min
2. [6:01pm] `user2` attempts to buy a ticket and sees the event is SOLD OUT
3. [6:30pm] `user1` fails or declines to complete their Stripe checkout
4. [6:30pm] Meet Near Me API receives the webhook and increments / restores the `inventory` from `0` => `1`
5. [6:31pm] `user2` and `user3` visit the website, and to `user2`'s surprise, there is a ticket available when the site had previously said "SOLD OUT"
6. [6:32pm] `user3` clicks "Checkout" before `user2`, decrementing the quantity optimistically again
7. [6:33pm] `user2` clicks "Checkout" but gets an error because the `inventory` is now `0` again, since they lost an unseen "race" against `user3`

I'm ok with this UX, what we can do is use websockets here to show `Live Quantity` and smooth this over. For now, I'm assuming there isn't tremendous competition for ticket sales among our B2B / smaller customers. If this changes, we will need the websocket implementation to show LIVE ticket availability

brianfeister:
This feature handles the (currently `PENDING`) status and resolves it via an API endpoint called by Stripe Webhook

The stripe webhook will check both the `Purchasable` and also `PurchasesV2` database tables to verify the purchase and then rollback the (previously decremented) `Inventory` on each item in the `Purchasable.PurchasableItems` array.

We introduce our first DB table deprecation + migration as well, because of a breaking schema change and a problem with the `/purchases` API which did not allow a user to purchase tickets multiple times for the same event) database tables

Fixes #201


Brandon-G-Tripp (approved):
This looks good to ship. Took me extra time with all of the changes in whitespace lit up like a Christmas tree. Ship it!

PR #205, opened 2024-11-01
Feature/marqo migration script

Brandon-G-Tripp:
# Pull Request: Marqo Migration Script 

## Description

This pull request introduces a Marqo migration script that allows migrations with the optional drop in "transformer". Key changes include the implementation of a GitHub workflow, batch processing with Go routines, error handling, and timestamp-based index naming.

## Features

  - Added GitHub workflow to execute the Marqo migration script.
  - Implemented batch processing using Go routines for efficient event item handling.
  - Introduced schema management in the migrator struct to allow better control over field protection.
  - Appended timestamp suffixes to new index names for better versioning and creation of new index during previous index teardown.
  - Enabled diagnostic testing for timezone storage issues in UTC.
  - Manual field handling, including proper cleaning of protected fields and timestamp corrections.
  - Updated URL construction logic for API calls, ensuring proper injection of new index names into schemas.
  - Logging to facilitate debugging during migrations.

## Summary by CodeRabbit

- **New Features**
	- Introduced a GitHub Actions workflow for manual migration of the Marqo application, allowing users to specify the environment, transformers, and batch size.
	- Added a command-line tool for migrating data to a Marqo database, supporting schema loading and error handling.
	- Implemented a structured JSON schema for database migration, detailing data types and properties.
	- Added transformer functions for data manipulation, including timestamp adjustments and name modifications.
	- Enhanced the workflow with improved debugging and error handling steps.
	- Introduced a new step for debugging directory structure in the workflow.
	- Updated the Marqo API client for better index and document management.

- **Bug Fixes**
	- Enhanced error handling throughout the migration process to ensure robust operation.

- **Documentation**
	- New schema file introduced to define structured data models for migration purposes.


internal/database/marqo/migration/client.go
```diff
@@ -0,0 +1,371 @@
...
+	Created string `json:"Created"`
+}
+
+type ListIndexesResponse struct {
+	Results []IndexInfo `json:"results"`
+}
+
+func NewMarqoClient(baseURL, apiKey string) *MarqoClient {
+	// Ensure baseURL ends with /api/v2
+	baseURL = strings.TrimRight(baseURL, "/")
```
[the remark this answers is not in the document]
brianfeister: I have the same question @Brandon-G-Tripp why `TrimRight` if we know the right side value is `/`? And is `/api/v2` part of `baseURL`?

internal/database/marqo/migration/client.go
```diff
@@ -0,0 +1,371 @@
...
+
+	return &MarqoClient{
+		baseURL: baseURL,
+		apiKey:  apiKey,
+		client:  &http.Client{},
```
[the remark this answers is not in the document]
brianfeister: @Brandon-G-Tripp this is actually a really good suggestion in a headless job like this so that we don't have hung process weirdness. Doesn't make as much sense in the webapp but here in a pipeline job it does

internal/database/marqo/migration/client.go
```diff
@@ -0,0 +1,371 @@
...
+		req.NumberOfInferences = 1
+	}
+
+	body, err := json.Marshal(req)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal request: %w", err)
+	}
+
+	fmt.Printf("Request body:\n%s\n", string(body))
+
```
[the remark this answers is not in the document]
brianfeister: I guess this is OK, though I'm not sure it's easy to 100% guarantee that this will never contain secrets

internal/database/marqo/migration/client.go
```diff
@@ -0,0 +1,371 @@
...
+		return "", fmt.Errorf("failed waiting for index: %w", err)
+	}
+
+	return endpoint, nil
+}
+
+// Helper function to create a structured index request from schema
+func CreateIndexRequestFromSchema(schema map[string]interface{}) (*CreateIndexRequest, error) {
+	data, err := json.Marshal(schema)
+	if err != nil {
```
brianfeister, line 147: Feels like we want an actual struct / type here to ensure we're passing valid schema. What do you think?

internal/database/marqo/migration/client.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+// Add these methods to MarqoClient
+
+func (c *MarqoClient) Search(indexName string, query string, offset, limit int) ([]map[string]interface{}, error) {
+	url := fmt.Sprintf("%s/indexes/%s/search", c.baseURL, indexName)
+	fmt.Printf("URL: %v", url)
+
+	requestBody := map[string]interface{}{
+		"q": "*", // use wildcard or empty query to retrieve all documents
```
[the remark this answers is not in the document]
brianfeister: @Brandon-G-Tripp I think this is hard-coded to `*` because we always assume "migration" (e.g. ALL documents), but Code Rabbit is right, if that's the case `query` should not be a function argument

internal/database/marqo/migration/client.go
```diff
@@ -0,0 +1,371 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read response body: %w", err)
+	}
+
+	fmt.Printf("Upsert response status: %d\n", resp.StatusCode)
+	fmt.Printf("Upsert response body: %s\n", string(bodyBytes))
+
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed to upsert documents: status=%d body=%s", resp.StatusCode, string(body))
```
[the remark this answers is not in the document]
brianfeister: Yeah maybe say "documents request body" or something here

internal/database/marqo/migration/migrate.go
```diff
@@ -0,0 +1,190 @@
...
+
+        if len(docs) == 0 {
+            break
+        }
+
+		// Send batch to worker
+		batchChan <- docs
+		activeBatches++
+		offset += m.batchSize
+
```
brianfeister, line 162: @Brandon-G-Tripp indentation is pretty whacky in this file. Would be good to get your IDE auto-formatting. It seems so small, but I always feel bad for all of the big diffs created by my changes prior to auto-formatting working correctly

internal/database/marqo/migration/migrate.go
```diff
@@ -0,0 +1,190 @@
...
+		if activeBatches == workers {
+			result := <-resultChan
+			if result.err != nil {
+				close(batchChan)
+				return result.err
+			}
+			totalMigrated += result.count
+			activeBatches--
+			fmt.Printf("Migrated and transformed %d documents\n", totalMigrated)
+		}
```
brianfeister, line 172: I had to look up what this was doing, but pretty impressive how parallelism is multi-threading is baked into go like this. It reminds me that if we ever do end up making TWO requests on the server for rendering a page, that we should use goroutines rather than just doubling our main thread blocking time 💡 

Pretty cool how it also handles race conditions with multiple out of sync resolutions of the batched goroutines resolving in different orders and sending back to the channel via [semaphore pattern](https://en.wikipedia.org/wiki/Semaphore_(programming)) I think that's the first time I'fe had a chance to write that word rather than just read it

internal/database/marqo/migration/schema.json
```diff
@@ -0,0 +1,208 @@
...
+  ],
+  "allFields": [
+    {
+      "name": "name_description_address",
+      "features": [],
+      "dependentFields": {
+        "name": 0.3,
+        "eventOwnerName": 0.1,
+        "description": 0.5,
+        "address": 0.2
```
brianfeister, line 29: when you're ready, we want to kick off a migration that adjusts these weights so that `name` is the significant leader per Tom's comments when we looked at this last

brianfeister (approved):
Nice touch with the "timestamp diagnostic" in the transformer output over here. Very exciting 👌 

https://github.com/meetnearme/api/actions/runs/11939067081/job/33278760458

I think I'd like to better understand how the transformer architecture is "pluggable" but also, there's a limit to how much I can understand your side of things, just like you're limited in ability to review my side.

Trusting you on this one. 🏆 🚀

internal/database/marqo/migration/schema.json
```diff
@@ -0,0 +1,208 @@
...
+        "filter"
+      ]
+    },
+    {
+      "name": "long",
+      "type": "double",
+      "features": [
+        "filter"
+      ]
+    },
```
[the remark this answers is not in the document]
brianfeister: despite the `schema.json` name @coderabbitai, this one is a specific schema where `validation` is not a valid key

internal/database/marqo/migration/client.go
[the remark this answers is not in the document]
Brandon-G-Tripp: @brianfeister This is not strictly needed but trimming it here makes all of the statements below much clearer when using Sprintf. For example. 

We can use this: 
	
url := fmt.Sprintf("%s/indexes/%s/search", c.baseURL, indexName)

instead of this: 
      
url := fmt.Sprintf("%sindexes/%s/search", c.baseURL, indexName)

Where we mash the formatting verb "%s" into the next part of the slug and make it harder to read.

internal/database/marqo/migration/migrate.go
Brandon-G-Tripp: On it.

internal/database/marqo/migration/migrate.go
Brandon-G-Tripp: Yes we should definitely keep this in mind. I am still getting comfortable with the routines and channels but I have a feeling we will find a ton of use for them with all of our server side rendering.

internal/database/marqo/migration/client.go
Brandon-G-Tripp: @brianfeister This should be addressed on the newest commit with a schema type in Go and a validate function. In the client.go file. 

Schema type at line 30 
Index Creation at line 183 or so.

internal/database/marqo/migration/transformers.go
```diff
@@ -0,0 +1,132 @@
...
+type TimestampInfo struct {
+	OriginalTimestamp int64
+	UTCTime string
+	LocalTime string
+	Timezone string
+}
+
+// TransformerRegistry holds all available transformers
+var TransformerRegistry = map[string]TransformFunc{
+	"name_transformer": func(doc map[string]interface{}) (map[string]interface{}, error) {
```
brianfeister, line 21: @Brandon-G-Tripp nice. A map of functions. This was the transformer logic I wasn't sure about. Looks good

brianfeister (approved):
Ship it 🚢  ... I read the transformer logic now. Cool to see this in GH actions now. LGTM

image

PR #229, opened 2024-12-14
[release] Promote partial event admin work + default home page filtering `dev` => `prod`

brianfeister:


brianfeister:
Test coverage is at `44.1%` (required is `45%`) will fix this later, need to get the release out today

PR #235, opened 2024-12-16
[fix] Fix broken CSS hashing

brianfeister:
Modify CSS hashing strategy to change the name of the file (rather than just a querystring value) and handle the filename reference a la `styles.<hash>.css` in `layout.templ` rather than the former `?h=<hash>`strategy which doesn't break cloudflare / browser caching

## Summary by CodeRabbit

- **New Features**
	- Updated stylesheet linking to use versioned filenames for improved asset management.
	- Introduced a mechanism for managing temporary CSS files during updates.

- **Bug Fixes**
	- Clarified warning comments regarding automatic updates of the stylesheet filename.

- **Documentation**
	- Enhanced comments for better understanding of asset management processes.

PR #241, opened 2024-12-18
Fixes #235 which failed to account for the new hash value in the comparison check

brianfeister:
…

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved logic for updating CSS file names based on hash values, reducing unnecessary updates.
	- Enhanced error handling for CSS update processes with warning logging.

- **Chores**
	- Refined the mechanism for extracting and comparing hash values in CSS filenames.


brianfeister:
Fixes bug with #218

PR #242, opened 2024-12-18
[fix] Post CSS hashing watcher fix

brianfeister:
This one fixes a gap to cover initial file creation when one doesn't exist (CI ignores these and they will always not exist in CI runs)

@coderabbitai ignore


brianfeister:

brianfeister:
Fixes bug with #218

PR #243, opened 2024-12-18
[fix] PostCSS watcher fix - guard against infinite loop triggered from changes to `layout.templ`

brianfeister:
Store the hashed value of `layout.templ`'s content to guard against infinite loop in PostCSS watcher


brianfeister:

PR #257, opened 2025-01-08
[release] Event Admin Pt 1

brianfeister:
This mostly completes the event admin work, though there is more to do.

Release contents:
* #252


brianfeister:
🚀

PR #275, opened 2025-01-17
created ability to check if geolocation is available in browser

sweedarbk:
This is just the ability to check if geolocation is available in the browser. Created this branch and pushed to save my work so I can do a git pull and get up to date.

### DIFF
--- a/functions/gateway/templates/pages/home.templ
+++ b/functions/gateway/templates/pages/home.templ
@@ -8,6 +8,32 @@ import (
 	"strconv"
 )
 
+templ geolocate() {
+	<script>
+	function locationData() {
+        return {
+            latitude: null,
+            longitude: null,
+            getLocation() {
+                if (navigator.geolocation) {
+                    navigator.geolocation.getCurrentPosition(
+                        (position) => {
+                            this.latitude = position.coords.latitude;
+                            this.longitude = position.coords.longitude;
+                        },
+                        (error) => {
+                            console.error("Error getting location:", error);
+                        }
+                    );
+                } else {
+                    console.error("Geolocation is not supported by this browser.");
+                }
+            },
+        };
+    }
+	</script>
+}
+
 script navigateToSlideHandler(index int) {
 		const currentIndex = parseInt(document.getElementById('current-slide-index').value);
 		const newIndex = currentIndex + index;
@@ -145,6 +171,7 @@ func GetCityCountryFromLocation(cfLocation helpers.CdnLocation, latStr, lonStr, 
 	}
 	return "Location: Please Select"
 }
+
 
 func GetCitiesAsJsonStr() string {
 	cityData := make([]map[string]interface{}, 0, 200)
@@ -470,6 +497,36 @@ templ HomePage(events []types.Event, pageUser *helpers.UserSearchResult, cfLocat
 			}
 		}
 
+function locationData() {
+	if ('geolocation' in navigator) {
+  console.log('Geolocation is Available');
+} else {
+  console.log('Geolocation is NOT Available');
+}
+        return {
+            latitude: null,
+            longitude: null,
+            getLocation() {
+                if (navigator.geolocation) {
+                    navigator.geolocation.getCurrentPosition(
+                        (position) => {
+                            this.latitude = position.coords.latitude;
+                            this.longitude = position.coords.longitude;
+                        },
+                        (error) => {
+                            console.error("Error getting location:", error);
+                        }
+                    );
+                } else {
+                    console.error("Geolocation is not supported by this browser.");
+                }
+            },
+        };
+    }
+
+	window.addEventListener('load', function() {
+    locationData();
+});
 		function getLocationSearchData() {
 			return {
 				allOptions: JSON.parse(
--- a/functions/gateway/templates/pages/layout.templ
+++ b/functions/gateway/templates/pages/layout.templ
@@ -23,7 +23,7 @@ templ Layout(sitePage helpers.SitePage, userInfo helpers.UserInfo, pageContent t
 			<link href="https://fonts.googleapis.com/css2?family=Outfit:wght@400&family=Ubuntu+Mono:ital,wght@0,400;0,700;1,400;1,700&family=Anton&family=Unbounded:wght@900&display=swap" rel="stylesheet"/>
 			// 🚨 WARNING 🚨 This filename is automatically updated by PostCSS
 			// ✅ DO commit it to version control whenever you see it change
-			<link rel="stylesheet" href={ templ.EscapeString(os.Getenv("STATIC_BASE_URL") + "/assets/styles.28d1fff7.css") }/>
+			<link rel="stylesheet" href={ templ.EscapeString(os.Getenv("STATIC_BASE_URL") + "/assets/styles.7281a955.css") }/>
 			<meta name="viewport" content="width=device-width, initial-scale=1"/>
 			<script src="https://unpkg.com/htmx.org@1.9.12"></script>
 			<script src="https://unpkg.com/htmx.org@1.9.12/dist/ext/json-enc.js"></script>
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,6 @@
         "node-watch": "^0.7.4",
         "sst": "2.39.5",
         "tailwindcss": "^3.4.9",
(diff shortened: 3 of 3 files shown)


functions/gateway/templates/pages/home.templ
```diff
@@ -8,6 +8,32 @@
...
+                        }
+                    );
+                } else {
+                    console.error("Geolocation is not supported by this browser.");
+                }
+            },
+        };
+    }
+	</script>
+}
```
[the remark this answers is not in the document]
brianfeister: @sweedarbk in this case @coderabbitai is correct, you have duplicated the JS logic here. We can delete this one and keep the one that adds to the existing `<script>` tag below

functions/gateway/templates/pages/home.templ
```diff
@@ -470,6 +497,36 @@ templ HomePage(events []types.Event, pageUser *helpers.UserSearchResult, cfLocat
 			}
 		}
 
+function locationData() {
+	if ('geolocation' in navigator) {
+  console.log('Geolocation is Available');
+} else {
+  console.log('Geolocation is NOT Available');
+}
```
brianfeister, line 505: This logic is duplicated below in the `else` block where you log `Geolocation is not supported by this browser`. If you want to affirm this, you could add a `console.log` in the `if` contained within `getLocation()` below

functions/gateway/templates/pages/home.templ
```diff
@@ -470,6 +497,36 @@ templ HomePage(events []types.Event, pageUser *helpers.UserSearchResult, cfLocat
...
+                        (error) => {
+                            console.error("Error getting location:", error);
+                        }
+                    );
+                } else {
+                    console.error("Geolocation is not supported by this browser.");
+                }
+            },
+        };
+    }
```
brianfeister, line 525: One thing that isn't currently happening in this JavaScript is hooking it up to Alpine state. In order to expose the properties to Alpine and call `latitude` and `longitude` within your UI from Alpine, it needs to be invoked via `x-data` tag.

Currently, we are doing this in `x-data="getHomeState()"`

Something that could be confusing right now is the question "how do we separate state so it's not all a big global mess?"

If you notice the current state of things, it's an accurate critique to say "this feels like it's all one big global scope and disorganized"

So, how do we fix this? We need to have a singular top-level Alpine scope but how do we componentize and isolate the child controllers and pass their data upstream?

I'm open to other solutions, but I recently tested this out with `templates/components/location_lookup.templ`

https://github.com/meetnearme/api/blob/develop/functions/gateway/templates/components/location_lookup.templ#L31-L33

The idea here is that the child uses `$dispatch` to message a `lat-updated` event on Alpine global scope to communicate the value has changed. Then, on the other end `getAddEditEventState()` (the top-level Alpine `getData()` scope for the page) listens for this `lat-updated` scope and updates the parent `formData.event.lat`

https://github.com/meetnearme/api/blob/develop/functions/gateway/templates/pages/event_add_edit.templ#L16-L18

brianfeister: This allows us to have a cleaner separation of concerns by using a messaging bus to intentionally listen / opt-in / update reactive UI rather than everything being in a shared global namespace

Part of me thinks eventually we'll use TS for type checking, and inject that compiled code, but that's a hypothetical future. 

For now, we let Golang (which has a great type system) drive by attempting to inject certain strongly-typed data in `data-[my-attribute]=""` tags so that the JavaScript isn't "driving" from a typing perspective and Golang types are the singular source of truth

functions/gateway/templates/pages/home.templ
```diff
@@ -470,6 +497,36 @@
...
+
+	window.addEventListener('load', function() {
+    locationData();
+});
```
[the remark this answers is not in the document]
brianfeister: @sweedarbk @coderabbitai isn't far off here. Indeed, it follows the existing pattern to use the existing `init()` that invokes when Alpine bootstraps `x-data="getHomeState()` is executed. 

Let's try to improve that pattern to make things a bit less chaotic for you without adding **yet another** method to the growing `getHomeState()` Alpine `x-data` scope

But let's try to "componentize" / modularize your work. It can have a new `<script>` tag and even a new `.templ` template component. From there, you can handle the bootstrapping inside it's own `init()` block and follow the `$dispatch` pattern to message that upward to the parent Alpine scope in `getHomeState()` which can follow the pattern from `@lat-updated.window="formData.event.lat = $event.detail.value"` to listen for updates on the `window`

functions/gateway/templates/pages/home.templ
```diff
@@ -531,21 +572,18 @@ templ HomePage(events []types.Event, pageUser *helpers.UserSearchResult, cfLocat
...
-										var pushState = history.pushState;
-										history.pushState = function(state) {
-												if (typeof history.onpushstate == "function") {
-														history.onpushstate({state: state});
-												}
-												// whatever else you want to do
-												// maybe call onhashchange e.handler
-												return pushState.apply(history, arguments);
-										}
-								})(window.history);
```
brianfeister, line 548: @sweedarbk I think I might have given some bad direction on this one

Here in discord
https://discord.com/channels/1096208627192320030/1151217204126290011/1333616994552905860

I was talking about "what is this doing?", when in fact it was my code, I think? 

We don't want to remove this code (the opposite of what I said before)

This is how the whole mechanism of filtering works by binding `window.history` so that we're essentially listening for changes on the state, which include query param changes and drive the whole thing

Removing this will make all the other query params (beside `lat` and `lon`) stop working

functions/gateway/templates/pages/home.templ
```diff
@@ -470,6 +471,46 @@ templ HomePage(events []types.Event, pageUser *helpers.UserSearchResult, cfLocat
...
+                            this.longitude = position.coords.longitude;
+                            this.isLoading = false;
+							pushToQueryParams('lat', this.latitude);
+                            pushToQueryParams('lon', this.longitude);
+                        },
+                        (error) => {
+                            const messages = {
++                            PERMISSION_DENIED: 'Please enable location access to use this feature.',
++                            POSITION_UNAVAILABLE: 'Unable to determine your location.',
++                            TIMEOUT: 'Location request timed out.'
```
brianfeister, line 496: @sweedarbk did you check that these constants are indeed the full range of possible options passed in `error` for the `navigator.geolocation()` callback?

functions/gateway/templates/pages/home.templ
```diff
@@ -470,6 +471,46 @@ templ HomePage(events []types.Event, pageUser *helpers.UserSearchResult, cfLocat
...
+                    );
+                } else {
+                    console.log('Geolocation is NOT Available');
+                    this.error = 'Geolocation is not supported by your browser';
+                }
+            },
+        };
+    }
+
+	window.addEventListener('load', function() {
```
brianfeister, line 511: This isn't a **big** deal, more of a nitpick, but I'm using the `$dispatch` function to send 

So, in alpine the `window.load` event is handled more via Alpine's `init()` as seen down here

https://github.com/meetnearme/api/pull/275/files#diff-f6cab24e67d1ac12108b30a5728b89be96a8c42f9a97378516af3183548be53bR573

So I guess you would typically do this in it's own file (for the modularity we talked about)

```
templ GeolocationScript() {
  // this bootstraps and calls Alpine's init()
  <div 
    x-data="geolocationData()"
    // we don't need distinct lat / lon, since it will always be both or neither
    $watch('latLon', value => $dispatch('geolocation-updated', { value }));
  ></div>
  
  <script>
     function locationData() {
       return {
         init() {
           // navigator.geolocation call here
           // set `lat` and `lon` local scope variables in the geolocation success callback
         },
         latLon: {lat: null, lon: null},
       }
     }
  </script>
}
```

And then listen for the updates like this in `home.templ`

```
    <div 
      x-data="getHomeState()" 
      class="p-4" 
      @location-selected.window="handleLocationSelected"
      @geolocation-updated.window="handleGeolocationUpdate"
		>
```

and then also in `home.templ` add a method to the returned `getHomeState()` controller

```
		function getHomeState() {
			return {
			  // ... (keep what else is there)
			  handleLocationSelected() {
			  // handle setting the dispatched state values from the downstream component to the `getHomeState()` internal state
			  }

```

brianfeister:
Approaching this in a different way

PR #283, opened 2025-01-22
Fix wrangler deployments 

brianfeister:
per https://github.com/cloudflare/wrangler-action/issues/311


brianfeister:
@coderabbitai ignore

PR #293, opened 2025-02-06
[feature] subdomain fixes, non-auth enablement for competitions, swap to "event series" parent in event list

brianfeister:


brianfeister:
@coderabbitai pause

PR #292, opened 2025-02-05
Pre-launch bugfix festival 🎉 

brianfeister:
* Add leaderboard / rounds to community pages
* Fix competition edit permissions issue where admins couldn't create
* Increase max length for community / user page "About" (`500` chars => `750` chars)
* Add new route /api/competition-config/owner/{ user Id}` to fetch standings on community pages
* Add new CompetitionConfig GSI for `endDate` sorting by competitions that have a future end date
* initial pass (still broken) at timepickers in competition admin + wire up types
* Fix bug in team user creation in zitadel
* Fix silent nil pointer error when loading user / community page if there is no user data  (was being swallowed by goroutine concurrency channel)
* Modify `helpers.GetDatetimePickerFormatted` to return unix time when timezone is optionally omitted
* Fix broken unit test for `TestGetUsersHandler`

image

## Summary by CodeRabbit

- **New Features**
  - Competition management pages now include new start and end time fields for scheduling.
  - Home page now dynamically retrieves and displays competition information.
  - Added functionality for retrieving competition configurations based on the primary owner's user ID.

- **Enhancements**
  - Profile pages support longer bios with an increased character limit.
  - Updated styling provides a refreshed visual experience.
  - Improved error handling and metadata processing for user and competition configurations.

PR #295, opened 2025-02-06
[release] 🚀 Big release: competitions module, new SST resources, community pages, & event series parent in main event stream

brianfeister:
Release contents

* #290 
* #291
* #292 
* #293 
* #294


brianfeister:
Test are passing despite ❌ , just need to get coverage threshold back up

PR #303, opened 2025-02-08
[fix] Allow empty competition rounds

brianfeister:
Previously a dynamo error would be thrown about "batch write update failed" when an empty array slice was passed to `UpdateCompetitionConfig` handler was called with empty array `[]`


brianfeister:
tests passing ✅ only coverage threshold is failing (will fix)

PR #304, opened 2025-02-08
[feature + fix] Add competition event parent association + fix event editor role bug + fix event series children in admin bug that hid past events

brianfeister:


brianfeister:
Tests are passing ✅ failure is just coverage, need to improve

PR #306, opened 2025-02-09
[feature] "Draft" Event Publish Status

brianfeister:
This PR introduces the ability to create "draft" / "unpublished" events or revert already published events to a "draft" / "unpublished" state

Additional details
* By default, "unpublished" events do not appear in search results / home page event list
* Non-editors will see `This event is unpublished`
* Editors can still see / edit / proofread unpublished events (if they're logged in)
* Button now shows `Save as Draft` when `Publish Status` is set to `Draft`
* Adds test coverage for these scenarios

Screenshots
image

image
image


brianfeister:
Tests passing ✅ just coverage threshold failure, which we have on our radar to fix

PR #328, opened 2025-02-13
[merge back] pull in hotfixes from `develop`

brianfeister:
## Summary by CodeRabbit

- **Refactor**
	- Updated the competition interface so that the event link now directs users to a public event view instead of an admin edit page.
	- Simplified the event editing page by removing the event source header and streamlining related display logic for series events.


brianfeister:
@coderabbitai ignore

PR #332, opened 2025-02-13
[fix] Rounds data cross-pollution `event admin` <> `competition admin`

### LINKED ISSUE #309: [bug] Cross data pollution in Competition Admin <> Event Admin
When loading data in competition admin and event admin respectively, we fetch data from the server for `CompetitionRounds`. The problem with this is that when we WRITE data, we do this in way that can cause conflict (last-in-first-out problems, and also stale data being upserted to the database)

Right now, we allow the following flow

1. Override Parent Event >> Add Competition

Image

2. Select a competition from dialog
3. Select rounds from that competition
Image

4. At this point, the rounds data is saved locally, and clicking `Add Selected Rounds` will make a direct PUT call. The problem here is, the `CompetitionConfig` can be out of sync where it's possible for the `CompetitionRounds` data to have been updated AFTER the time the rounds were loaded in this dialog

We want to instead, attach `rounds` to the `this.formData.event` and `this.formData.eventChildren` Alpine controller data, depending on which button was clicked to open the dialog

Once this is done, we need to update the `/api/event-reg-purch` endpoint to make a `PATCH` call to the `CompetitionRounds` table **but only update the `competitionRound.eventId` field**

The same update needs to happen from the `/admin/competition` UI, which is to say, explicitly **OMIT** the `eventId` from the payload, because it could have been changed on the `/admin/event` page, which means sending the payload will overwrite / negate the configuration work the user just did on `/admin/event`

This means we need to explicitly PRUNE the `.eventId` from the competition config payload

Brandon-G-Tripp:
This PR closes #309

It removes the direct call to the rounds api from "Add Selected Rounds" and attaches round updates to the Publish payload. It passes in the new associated eventId and only updates the eventId from the event admin page. It also modifies the competitioinConfig Update function to call BatchPatchRounds to only add eventId if it is the placeholder 'fake-event' etc. 

## Summary by CodeRabbit

- **New Features**
  - Event registration and purchasing now support updates for competition rounds.
  - Competition management has been enhanced for smoother team assignments, including streamlined addition and removal of competitors.
  - Event pages now manage pending round updates with a safeguard against publishing too many rounds.
  - The user selection component has been improved for dynamic handling during removals.
  - When creating new competitions, the creator is automatically designated as the primary owner.

### DIFF
--- a/functions/gateway/templates/pages/event_add_edit.templ
+++ b/functions/gateway/templates/pages/event_add_edit.templ
@@ -493,8 +493,22 @@ templ AddOrEditEventPage(pageObj helpers.SitePage, event types.Event, isEditor b
 												/>
 											}
 										</div>
-										<template x-if="child.competitionConfigId">
+										<div x-text="JSON.stringify(child)"></div>
+										<template x-if="child && child.competitionConfigId">
 											<div class="form-control">
+												<div
+													x-init="
+                          console.log('Child event debug:', {
+                              childId: child.id,
+                              hasConfigId: !!child.competitionConfigId,
+                              configId: child.competitionConfigId,
+                              roundsLoaded: competitionRoundsLoaded,
+                              hasRounds: !!child.competitionRounds,
+                              roundsCount: child.competitionRounds?.length,
+                              rounds: child.competitionRounds
+                            })
+                        "
+												></div>
 												<label class="label">Competition Config</label>
 												<a class="link link-primary" :href="'/admin/competition/' + child.competitionConfigId + '/edit' ">
 													View Competition Config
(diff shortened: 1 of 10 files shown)


functions/gateway/templates/pages/event_add_edit.templ
```diff
@@ -1523,88 +1524,74 @@ templ AddOrEditEventPage(pageObj helpers.SitePage, event types.Event, isEditor b
...
+						this.saveReqInFlight = true;
+						try {
+								// Create array of round updates for pending updates (includes eventId)
+								const roundsUpdatePayload = this.competitionModalRounds.filter(round =>
+										this.competitionModalRoundSelections.includes(round.roundNumber.toString())
+								).map(round => ({
+										...round,
+										eventId: this.competitionModalEventSelection.id
+								}));
+
```
brianfeister, line 1537: @Brandon-G-Tripp I think the functionality might be misunderstood here. The `this.competitionModalRoundsSelections` used to assume the following UX:

1) User opens modal
2) Select rounds
3) Immediately update them when clicking `Save` in the dialog

I believe with your addition of `comptitionRounds` as a payload on `/api/event-reg-purch` endpoint, that your change here does not allow a user to do the following:

1) Add 3 rounds to Event A
2) Add 3 rounds to Event B
3) Batch save all. 

Because you're leaning on `this.competitionModalRoundsSelections` but that array doesn't include round associate data right now... or does it? I'm trying to understand what's happening in your UI logic

Brandon-G-Tripp: Well the logic is pulling off the rounds in the function and saving it to local state on line 1562. But you are right. I believe I need to append the round updates at that place so they are saved for when the modal is opened and rounds are selected multiple times.

PR #335, opened 2025-02-18
[feature] Add competitions active indicator + Fix competition owner save bug

brianfeister:


brianfeister:
Test passing ✅ (coverage threshold still needs help)

PR #341, opened 2025-02-24
[fix] Competition Round Association

brianfeister:
This PR is a huge fix for event admin, this is officially WAY to complex + Alpine is now getting in the way with nested `<template x-if="">` creating nested object.observe stateful issues with reactivity... Mostly Alpine has been great, but this is the first time I wished for React and this part of the app is now outrageously complex and problematic...

Fixes:
* hide event links from competition admin when they are empty placeholder ID
* huge refactor to competition association (addition) logic to create a queue of "pending updates" to competitionRounds and reconcile them with iterative stateful changes
* huge refactor to deletion logic for "soft delete" which is all committed in a single save action
* dropped <template x-if=""> in places where the nested templates were causing issues for Alpine reactivity (nested elements wouldn't appear as they should)
* auto-delete `competitionConfig` association when soft-deleting (marked with empty placeholder ID)
* Make both event parent and event children competition association work

## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted conditions to ensure event links display only when valid, preventing unintended visibility.
  
- **Refactor**
  - Streamlined competition rounds and configuration display logic for a more consistent and responsive event management interface.

### DIFF
--- a/functions/gateway/templates/pages/competition_add_edit.templ
+++ b/functions/gateway/templates/pages/competition_add_edit.templ
@@ -305,7 +305,7 @@ templ AddOrEditCompetitionPage(pageObj helpers.SitePage, competition types.Compe
 											</svg>
 										</button>
 									</div>
-									<template x-if="round && round?.eventId">
+									<template x-if="round && round?.eventId && round?.eventId !== emptyRoundEventId ">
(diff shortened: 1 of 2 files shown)


brianfeister:
@coderabbitai ignore

PR #343, opened 2025-02-25
[fix] Don't send entire `competitionRound` payload from event admin

brianfeister:
sending too much data causes status like `isVotingOpen` to get overwritten. In the previous release to dev, saving from Event Admin if loaded in a stale state could overwrite `competitionRound` keys that have an intentionally high change rate like `isVotingOpen` or `competitor(A|B)Score`

## Summary by CodeRabbit

- **Refactor**
  - Enhanced the event creation and editing experience by consolidating round details into a unified view. This update streamlines event scheduling workflows by automatically integrating relevant round information, ensuring that only essential details are processed during event publishing for a smoother overall experience.


brianfeister:
@coderabbitai ignore

PR #346, opened 2025-02-25
[fix] Migrate from `access_token` / `refresh_token` ==> `mnm_access_token` /`mnm_refresh_token`

brianfeister:
This fixes an issue where users could get stuck in a redirect loop due to the invalid legacy tokens prior to the subdomain / logout fix here:

* #338
* #339
* #340

⚠️  NOTE ⚠️  Some places in code still show `access_token`/`refresh_token` because the underlying exchange with Zitadel assumes that keyname


brianfeister:
@coderabbitai ignore

PR #368, opened 2025-04-09
[fix] Location search / geolocation / filter nav flyout

brianfeister:
* Fix UI reactivity in `Showing events within...` section
* Fix `.menu` CSS class pollution of location search combobox
* UI tweaks and cleanup


brianfeister:
@coderabbitai ignore

PR #376, opened 2025-04-12
[fix] Event time range query

brianfeister:
This PR fixes a bug where "ongoing events" (those with a `startTime` in the past, but an `endTime` in the future) were not shown in search results


brianfeister:
Closing as this includes an unintended commit that should not go to `develop` yet

PR #380, opened 2025-04-15
fix geocoordinate fallback behavior so that cloudflare city takes precedence

brianfeister:
Before this the (not very human friendly / readable `Geocoordinates: 35.35, -75.23`) would show in the UI like this

image

We want human readable to always take precedence in these situations, so the new priority order is:

1) `location` query param
2) cloudflare location city / country
3) geocoordinates

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved location display logic to prioritize showing city and country code when available.


brianfeister:
@coderabitai ignore

PR #391, opened 2025-05-07
[fix] Require `endTime` for all competitions

brianfeister:
Since Dynamo indexes on `endTime` for competition configs, we need to enforce that all competitions to be saved must have an end time


brianfeister:
@coderabbitai ignore

PR #396, opened 2025-06-04
[fix] Revert #394 (doesn't fix the issue) and properly wire up new `ARE_API_CLIENT_BASE_URL` env var

brianfeister:
## Summary by CodeRabbit

- **Chores**
	- Added a new environment variable for Cloudflare API client base URL across environment files and deployment workflows.
- **Refactor**
	- Updated event processing logic to always handle events directly, removing batching and parallel processing.


brianfeister:
I'm merging this with failing tests because I believe the tests are failing due to a missing env var `CLOUDFLARE_API_CLIENT_BASE_URL` which (chicken and egg) needs to merge to `develop` first before the `deploy-dev.yml` picks up the change

Will reference a passing build link here afterwards

